Issue #3, opened 2015-01-27
Layer’s transform-derived properties


saniul:
The Layer properties that abstract away the layer’s transform matrix need to be re-thought. Currently they can easily go out of sync with the layer’s transform. 

This can easily occur, e.g. when animating `rotationRadians` the dynamic animator will animate the layer’s transform directly and after the animation is done, if the layer’s `rotationRadians` property is read it will not reflect the truth.

What do you think about using KVC to extract and set the underlying CALayer’s `transform` property? So getting the rotation would be done using `layer.setValue(newValue, forKeyPath:”transform.rotation.z")` and getting `layer.valueForKeyPath(“transform.rotation.z”)`.

PR #4, opened 2015-01-27
RotationGesture + fixed scaleX, scaleY, rotationRadians going out of sync with layer.transform

saniul:
Implemented a rotation gesture as well as proposed changes discussed in #3

### DIFF
--- a/Prototope/Gesture.swift
+++ b/Prototope/Gesture.swift
@@ -314,7 +314,6 @@ public class RotationGesture: GestureType {
             let rotationGesture = gestureRecognizer as UIRotationGestureRecognizer
             switch rotationGesture.state {
             case .Began:
-                // Reset the gesture to record rotation relative to the starting rotation
                 let centroidWindowLocation = rotationGesture.locationInView(nil)
                 
                 struct IDState { static var nextCentroidSequenceID = 0 }
--- a/Prototope/Gesture.swift
+++ b/Prototope/Gesture.swift
@@ -196,7 +196,7 @@ public class PanGesture: GestureType {
 	}
 }
 
-/** A rotation gesture represents the state of a rotation gesture at a single point in time */
+/** A rotation sample represents the state of a rotation gesture at a single point in time */
 public struct RotationSample: SampleType {
     public let rotationRadians: Double
     public let velocity: Double
--- a/Prototope/Gesture.swift
+++ b/Prototope/Gesture.swift
@@ -369,39 +369,39 @@ public func +<Seq: SampleSequenceType>(a: Seq, b: Seq) -> Seq {
 public struct SampleSequence<S: SampleType, I: Printable> : SampleSequenceType {
     typealias Sample = S
     typealias ID = I
-    
+
     /** Samples ordered by arrival time. */
-    public let samples: [S]
+    public let samples: [Sample]
     
     /** An identifier that can be used to distinguish this sequence from e.g. other
     sequences that might be proceeding simultaneously. You might think of it as
     a "finger identifier". */
-    public var id: I
+    public var id: ID
     
     /** The first sample. */
-    public var firstSample: S! {
+    public var firstSample: Sample! {
         return samples.first
     }
     
     /** The next-to-last sample (if one exists). */
-    public var previousSample: S? {
+    public var previousSample: Sample? {
         let index = samples.count - 2
         return index >= 0 ? samples[index] : nil
     }
     
     /** The most recent sample. */
-    public var currentSample: S! {
+    public var currentSample: Sample! {
         return samples.last
     }
     
-    public init(samples: [S], id: I) {
+    public init(samples: [Sample], id: ID) {
         precondition(samples.count >= 0)
         self.samples = samples
         self.id = id
     }
     
     /** Create a new sequence by adding a sample onto the end of the sample list. */
-    public func sampleSequenceByAppendingSample(sample: S) -> SampleSequence<S, I> {
+    public func sampleSequenceByAppendingSample(sample: Sample) -> SampleSequence<Sample, ID> {
         return SampleSequence(samples: samples + [sample], id: id)
     }
     
@@ -420,7 +420,7 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
     typealias ID = I
     
     /** Inner sequence */
-    private let sequence: SampleSequence<TouchSample, I>
+    private let sequence: SampleSequence<TouchSample, ID>
     
     /** Touch samples ordered by arrival time. */
     public var samples: [TouchSample] {
@@ -430,7 +430,7 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
     /** An identifier that can be used to distinguish this touch sequence from e.g. other
     touch sequences that might be proceeding simultaneously. You might think of it as
     a "finger identifier". */
-    public var id: I {
+    public var id: ID {
         return sequence.id
     }
     
@@ -449,7 +449,7 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
         return sequence.currentSample
     }
     
-    public init(samples: [TouchSample], id: I) {
+    public init(samples: [TouchSample], id: ID) {
         self.sequence = SampleSequence(samples: samples, id: id)
     }
     
@@ -479,7 +479,7 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
     }
     
     /** Create a new touch sequence by adding a sample onto the end of the sample list. */
-    public func sampleSequenceByAppendingSample(sample: TouchSample) -> TouchSequence<I> {
+    public func sampleSequenceByAppendingSample(sample: TouchSample) -> TouchSequence<ID> {
         return TouchSequence(samples: samples + [sample], id: id)
     }
     
--- a/Prototope/Gesture.swift
+++ b/Prototope/Gesture.swift
@@ -11,7 +11,7 @@ import UIKit
 // MARK: - Touches
 
 /** Represents the state of a touch at a particular time. */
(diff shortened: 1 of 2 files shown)


saniul:
Not sure how useful the centroid path tracking is for the rotation gesture if we’re not recording the rotation itself over time. Didn’t do anything special here to keep it simple.

Prototope/Layer.swift
```diff
@@ -253,13 +258,23 @@ public class Layer: Equatable, Hashable {
 	}
 
 	/** The scaling factor of the layer along the x dimension. Defaults to 1. */
-	public var scaleX: Double = 1 {
-		didSet { updateTransform() }
+	public var scaleX: Double {
+        get {
+            return layer.valueForKeyPath("transform.scale.x") as Double
```
NachoSoto: I wonder if getting these values from the presentation layer would be more what we want? :)

saniul: Thanks for having a look at this!

I feel like that’d be an implementation detail that would easily leak when using this API. Say you set the rotation (by setting it at the `layer.transform` level) and then immediately query the rotation (this would extract it from `layer.presentationLayer.transform`) - you will get the old value.

Besides, it feels like the whole point of this was to abstract away some of the UIKit, CA framework weirdness behind an API that’s simpler, quicker to use and understand for a limited set of use cases. Exposing the CA implementation detail of layer/presentationLayer duality seems to go against this goal (to me). But let me know if I’m wrong :)

NachoSoto: I totally see what you mean, unfortunately I'm not sure we can hide UIKit completely. A parallel example to yours:
Say you animate a layer's rotation, and sometime during the animation you query the value, hoping to get the *current* rotation. You'd get the final value.

I'm not sure how we can effectively provide both values, while keeping the API simple and not exposing such CA details :/ But I haven't thought about this enough.

nsfmc:
as i understand it, i feel like the more natural approach would be to do something along the lines of allowing folks to add something like `Animator.currentAnimationHandler` where the `animationHandler` would receive the layer and the current value of the animated property? i would expect that *only* this sort of handler would have the 'live' value of the animation at any given moment and that any other queryable values might be stale or optimistic/future values. Similar to what's going on with Heartbeat.swift where you can attach a handler for each frame and modify properties then.

i dunno, does that make any sense?

saniul:
Having a handler that can fire on every frame of the animation sounds useful and is definitely something that we should consider adding regardless of this issue.

However, it still won’t solve the fact that users will have to *think* about whether they want to get the underlying model value (query the layer’s properties) or the “inflight” value (check if there’s an animation and if so - attach an animation handler).

Note: Am I wrong in thinking that this is only an issue with CoreAnimation’s animations? AFAIK, POP animates by setting the model value on every display frame, so they will reflect what’s presented on the screen.

Anyhow, this discussion warrants a separate Issue of its own. And I feel like the eventual solution should stem from the envisioned use case, as well as whether or not Prototope will continue to support classic UIView/CAAnimations, as opposed to relying exclusively on POP animations.

Prototope/Layer.swift
andymatuschak: @saniul: You've totally got it right. I haven't actually written this out, but I intend to deal with this dichotomy through the distinction between `layer.animators.x.target` and `layer.x`. i.e. the first is "like the model value" and the second is "like the presentation value."

Yeah, we still have this semi-vestigial `Layer.animateWithDuration` method which doesn't use the POP animators, but I really don't like that API anyway. Gotta merge it with the dynamic animators as a convenience or otherwise make it go through the same paths. Haven't thought it through.

Prototope/Gesture.swift
```diff
@@ -268,6 +268,79 @@ public class PanGesture: GestureType {
 	}
 }
 
+/** A rotation gesture recognizes a standard iOS rotation: it doesn't begin until the user's moved some number of points, then it tracks new touches coming and going over time as well as rotation relative to the beginning of the gesture and the current rotation velocity. It exposes simple access to the path of the center of all the touches and the rotation and rotation velocity values. */
```
andymatuschak: s/moved some number of points/rotated by some number of degrees

Prototope/Gesture.swift
```diff
@@ -268,6 +268,79 @@ public class PanGesture: GestureType {
...
+public class RotationGesture: GestureType {
+    /** The handler will be invoked as the gesture recognizes and updates; it's passed the gesture's current
+    phase (see ContinuousGesturePhase documentation), rotation relative to the state at the beginning 
+    of the gesture in radians, rotation velocity in radians per second and also a touch sequence representing the center of all the touches involved in the rotation gesture. */
+    public convenience init(_ handler: (phase: ContinuousGesturePhase, rotationRadians: Double, rotationVelocity: Double, centroidSequence: TouchSequence<Int>) -> ()) {
```
andymatuschak: Wish there were some good way to make defining this closure handler more friendly. This many positional arguments is rough!

One idea: what if we defined a `RotationGesture.State` `struct` subtype (with `phase`, `rotationRadians`, etc fields) and passed one of those? Then you could do lightweight things like this:

```
layer.gestures.append(RotationGesture { layer.rotationRadians = $0.rotationRadians })
```

Prototope/Gesture.swift
```diff
@@ -268,6 +268,79 @@ public class PanGesture: GestureType {
...
+        
+        init(actionHandler: (phase: ContinuousGesturePhase, rotationRadians: Double, rotationVelocity: Double, centroidSequence: TouchSequence<Int>) -> ()) {
+            self.actionHandler = actionHandler
+        }
+        
+        func handleGestureRecognizer(gestureRecognizer: UIGestureRecognizer) {
+            let rotationGesture = gestureRecognizer as UIRotationGestureRecognizer
+            switch rotationGesture.state {
+            case .Began:
+                // Reset the gesture to record rotation relative to the starting rotation
```
andymatuschak: Hm, but the following line doesn't actually reset the gesture if I understand correctly.

Prototope/Layer.swift
andymatuschak: (@nsfmc, does my comment address your comment too?)

andymatuschak:
Really cool that you did this, @saniul! Appreciate the contribution. :100: 

It seems like this API really wants to be able to emit a "rotation state sequence," in the same sense that the other APIs emit e.g. a touch sequence. I know from experience that I'd want to be able to refer to e.g. the current rotation radians vs. the previous rotation radians vs. the initial rotation radians—really, for the same reasons that you'd want to be able to do that with touch locations. What do you think?

Prototope/Gesture.swift
saniul: Yeah, I removed that comment in https://github.com/saniul/Prototope/commit/cedb32fe6dca19c82562ee4bebcf3e040f85ec91

saniul:
Totally agree, I didn’t implement recording the state sequence to keep things simple, but it does make a lot of sense.

One thing though, I can’t really reconcile [simplifying the handler closure’s arguments](https://github.com/Khan/Prototope/pull/4#discussion_r23699786) down to a single state `struct` with emitting a rotation state sequence. Looks like it’ll have to be something like `(ContinuousGesturePhase, GestureStateSequence<RotationGesture.State>, TouchSequence<Int>) -> ()`, unless we somehow unify `TouchSequence` and `GestureStateSequence` (which doesn’t seem right to me).

I’ll work on the changes later today.

andymatuschak:
What if:

```swift
struct SampleSequence<Sample> {
 ...
}

typealias TouchSequence = SampleSequence<TouchSample>

struct RotationSample {
  let rotationRadians: Double
  let centroid: TouchSample
  ...
}

typealias RotationSampleSequence = SampleSequence<RotationSample>
```

etc

saniul:
You’ll have to excuse my wariness for changing the existing TouchSequence API :) Yeah, seems like that is the way to go. I’ll have to extract the `currentVelocityInLayer` into function since extensions on specialized generic types aren’t allowed (related: https://twitter.com/indragie/status/560539062398750722)

andymatuschak:
Ooh, yeah. Hm hm. Maybe make TouchSequence a decorator on SampleSequence instead of making it a typealias, so you can keep that as a method?

Prototope/Layer.swift
nsfmc: :+1: @andymatuschak

saniul:
@andymatuschak How’s that https://github.com/saniul/Prototope/commit/4efa8ab01ee6917f4acf6e670ce38a5cc7c99172? Not sure if [this](https://github.com/saniul/Prototope/commit/4efa8ab01ee6917f4acf6e670ce38a5cc7c99172#diff-cb323b36c43bdd9204ab393c98b3ce44R418) is quite what you wanted me to do for TouchSequence.

saniul:
Any updates on this? Would you like me to implement it differently?

andymatuschak:
Very sorry for the delay—I got sidetracked by many other things! You've done an excellent job with that refactoring! Thanks so much.

saniul:
@andymatuschak No worries!

- [x] Added `rotationDegrees` and `velocityDegrees` to `RotationSample` and modified the `description` to `"<RotationSample: ⟳\(rotationDegrees)° ∂⟳\(velocityDegrees)°/s, \(rotationRadians)rad \(velocityRadians)rad/s, @\(centroid)>”`
- [x] Changed `SampleSequence`and `TouchSequence` internals to refer to the original typealias names to make the code more readable

saniul:
:boom: 

I’ll update the OhaiPrototype PR to point to the updated submodule in about an hour.

andymatuschak:
Woo! Thanks so much @saniul! Really appreciate the contribution :100:

nsfmc:
:heart:

PR #7, opened 2015-02-06
[FIN] Simultaneous Gesture recognition support

saniul:
This adds a function property on Gestures that lets the user of the API control whether it should be recognized simultaneously with that gesture.

### DIFF
--- a/Prototope/Gesture.swift
+++ b/Prototope/Gesture.swift
@@ -86,6 +86,7 @@ public class TapGesture: GestureType {
 		tapGestureRecognizer.cancelsTouchesInView = cancelsTouchesInLayer
 		tapGestureRecognizer.numberOfTapsRequired = numberOfTapsRequired
 		tapGestureRecognizer.numberOfTouchesRequired = numberOfTouchesRequired
+        shouldRecognizeSimultaneouslyWithGesture = { _ in return false }
 	}
 
 	deinit {
@@ -106,6 +107,12 @@ public class TapGesture: GestureType {
 
 	private let tapGestureRecognizer: UITapGestureRecognizer
 	private let tapGestureHandler: TapGestureHandler
+    
+    public var underlyingGestureRecognizer: UIGestureRecognizer {
+        return tapGestureRecognizer
+    }
+    
+    public var shouldRecognizeSimultaneouslyWithGesture: GestureType -> Bool
 
 	public weak var hostLayer: Layer? {
 		didSet { handleTransferOfGesture(tapGestureRecognizer, oldValue, hostLayer) }
@@ -151,10 +158,18 @@ public class PanGesture: GestureType {
 		panGestureRecognizer.cancelsTouchesInView = cancelsTouchesInLayer
 		panGestureRecognizer.minimumNumberOfTouches = minimumNumberOfTouches
 		panGestureRecognizer.maximumNumberOfTouches = maximumNumberOfTouches
+        
+        shouldRecognizeSimultaneouslyWithGesture = { _ in return false }
 	}
 
 	private let panGestureRecognizer: UIPanGestureRecognizer
 	private let panGestureHandler: PanGestureHandler
+    
+    public var underlyingGestureRecognizer: UIGestureRecognizer {
+        return panGestureRecognizer
+    }
+    
+    public var shouldRecognizeSimultaneouslyWithGesture: GestureType -> Bool
 
 	public weak var hostLayer: Layer? {
 		didSet { handleTransferOfGesture(panGestureRecognizer, oldValue, hostLayer) }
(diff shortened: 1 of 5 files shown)


saniul:
There’s a lot of repetition between the gestures which could probably be avoided with inheritance (generic struct with a <GestureRecognizerType: UIGestureRecognizer> constraint?), but I was guessing that the intention was to avoid it.

saniul:
Added bridging for the `shouldRecognizeSimultaneouslyWithGesture` function

PrototopeTestApp/JSTest.swift
```diff
@@ -16,5 +16,7 @@ func makeJSLayer() {
 		let lineNumber = value.objectForKeyedSubscript("line")
 		println("Exception on line \(lineNumber): \(value)")
 	}
-	println(context.evaluateScript("var layer = new Layer({parent: Layer.root}); Layer.animate({duration: 3.0, curve: AnimationCurve.EaseOut, animations: function() { layer.x = 400 }}); layer.gestures = [new PinchGesture({handler: function(phase, sequence) {console.log(sequence.currentSample.scale)}})]; layer.backgroundColor = new Color({red: 0.5, green: 0.7, blue: 0.1, alpha: 0.7}); layer.frame = new Rect({x: 75, y: 80, width: 400, height: 400}); layer.border = new Border({color: Color.black, width: 2}); layer.shadow = new Shadow({alpha: 1.0}); (new Sound({name: 'Glass'})).play()"))
+    
+    let script = NSString(contentsOfURL: NSBundle.mainBundle().URLForResource("JSTest", withExtension: "js")!, encoding: NSUTF8StringEncoding, error: nil)
```
andymatuschak: omg thank you for doing this

I was always just one notch too lazy to bother. :)

andymatuschak:
Impressed that you waded through bridging this to JS. *reading*

andymatuschak:
I… no longer remember why I didn't make a shared superclass. The one thing I felt strongly about was that the core gesture interface used by the outside world should be defined in a protocol; i.e. that you should be able to make a class which conforms to `GestureType` but which does not inherit from `class BuiltinGesture` (or whatever).

I do have a pretty strong aversion to inheritance by default, and I'll tolerate some degree of repetition in exchange for avoiding it, but I think we're getting close to that threshold. What do you think? Time to Extract Class?

Prototope/Gesture.swift
```diff
@@ -593,19 +633,63 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
...
 
-private func handleTransferOfGesture(gesture: UIGestureRecognizer, fromLayer: Layer?, toLayer: Layer?) {
+private func handleTransferOfGesture(gesture:GestureType, fromLayer: Layer?, toLayer: Layer?) {
+    let recognizer = gesture.underlyingGestureRecognizer
+    
+    switch (fromLayer, toLayer) {
+    case (.None, .Some):
+        gestureMap[recognizer] = gesture
+    case (.Some, .None):
+        gestureMap.removeValueForKey(recognizer)
```
andymatuschak: Glad you were able to centralize this.

Prototope/Gesture.swift
```diff
@@ -593,19 +633,63 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
...
-		fromLayer?.view.removeGestureRecognizer(gesture)
-		toLayer?.view.addGestureRecognizer(gesture)
+		fromLayer?.view.removeGestureRecognizer(recognizer)
+		toLayer?.view.addGestureRecognizer(recognizer)
 	}
 }
 
 public protocol _GestureType {
 	weak var hostLayer: Layer? { get nonmutating set }
+    var underlyingGestureRecognizer: UIGestureRecognizer { get }
```
andymatuschak: I think I was trying to avoid this too, out of a sense of idealism (that maybe you'd want to define a gesture that is not a `UIGestureRecognizer`), but it makes way more sense to be pragmatic here.

andymatuschak:
This is looking really great, @saniul. Shall I go ahead and merge, or would you like to Extract Class?

saniul:
Oops, looks like there was a merge conflict after the last batch of master merges. Fixed now.

Prototope/Gesture.swift
```diff
@@ -593,19 +633,63 @@ public struct TouchSequence<I: Printable> : SampleSequenceType {
...
+// MARK: Gesture-to-gesture interaction
+
+//Need to have a way to map Gestures to UIGestureRecognizers
+var gestureMap = [UIGestureRecognizer:GestureType]()
+
+func gestureForGestureRecognizer(gestureRecognizer: UIGestureRecognizer) -> GestureType? {
+    return gestureMap[gestureRecognizer]
+}
+
+@objc class GestureRecognizerBridge: NSObject, UIGestureRecognizerDelegate {
```
andymatuschak: Nicely done.

PrototopeJSBridge/GestureBridge.swift
```diff
@@ -400,6 +431,20 @@ public class ContinuousGesturePhaseBridge: NSObject, BridgeType {
 	public class func addToContext(context: JSContext) {
 		context.setObject(self, forKeyedSubscript: "RotationGesture")
 	}
+    
+    public var shouldRecognizeSimultaneouslyWithGesture: JSValue? {
```
andymatuschak: Looks like we have an opportunity to extract these implementations to a superclass, too, but can happen in a future revision.

andymatuschak:
Looking great. Thank you!

PR #8, opened 2015-02-06
[WIP] Decay animation support

saniul:
**This isn’t ready to merge, this is more like a sketch to discuss changing the animation API.**

Adds `DecayAnimation` and extracts the previous spring animation properties into `SpringAnimation`.
This PR also adds an `applyHandler`, which allows processing on every POP animation tick.

### DIFF
--- a/Prototope/Animation.swift
+++ b/Prototope/Animation.swift
@@ -74,6 +74,61 @@ public class LayerAnimatorStore {
 	}
 }
 
+public protocol AnimationKind {
+    func createInstance() -> POPPropertyAnimation
+    func updateAnimation(animation: POPPropertyAnimation)
+}
+
+public struct DecayAnimation<Target: AnimatorValueConvertible> : AnimationKind {
+	public var velocity: Target?
+
+    public init(velocity: Target? = nil) {
+        self.velocity = velocity
+    }
+    
+    public func createInstance() -> POPPropertyAnimation {
+        return POPDecayAnimation()
+    }
+    
+    public func updateAnimation(animation: POPPropertyAnimation) {
+        if let animation = animation as? POPDecayAnimation {
+            if let velocityValue: AnyObject = velocity?.toAnimatorValue() {
+                animation.velocity = velocityValue
+            }
+        }
+    }
+}
+
+public struct SpringAnimation<Target: AnimatorValueConvertible> : AnimationKind {
+	public var velocity: Target?
+
+	/** How quickly the animation resolves to the target value. Valid range from 0 to 20. */
+    public var springSpeed: Double = 4.0
+	/** How springily the animation resolves to the target value. Valid range from 0 to 20. */
+	public var springBounciness: Double = 12.0
+
+    public init(velocity: Target? = nil, springSpeed: Double = 4.0, springBounciness: Double = 12.0) {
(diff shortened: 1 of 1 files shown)


saniul:
AnimationKind should probably be aliased by the property type, but that would prevent us from using it as a variable type in `Animator` :|

It seems like this whole API should be re-thought anyway... E.g.:
* There’s currently no way to get the current animation velocity;
* Modifying the dynamic animation while it’s in-flight feels weird and wrong right now;

andymatuschak:
Thanks very much for kicking this off! Props for the initial labor.

>It seems like this whole API should be re-thought anyway... E.g.:
> * There’s currently no way to get the current animation velocity;
> * Modifying the dynamic animation while it’s in-flight feels weird and wrong right now;

I think I probably haven't communicated very clearly the intent of these dynamic animators, so they feel a little alien.

The idea is that rather than creating a one-off *animation*, you may want to create a persistent entity—an *animator*—which is responsible for continuously animating a particular property in a way that keeps position and velocity smooth. So the intended idiom with that API really *is* literally "changing the target of the animator." And given that the animator takes care of keeping velocity smooth, I'm not sure why it's necessary to get the instantaneous velocity—what would you do with it?

Given that context, do the design decisions make more sense?

---

Regarding the API, I do want to keep things a little lighter-weight than this. Clients are now burdened with either keeping the `AnimationKind` instance around to modify e.g. its target, or else they must typecast the `animationKind` property.

I have a lot of ideas about how to do this in a principled fashion, but all of them would make the API heavier. I think we should do an unprincipled thing because this is a prototyping framework. :)

More specifically, I think we should just straight-up expose the union of the relevant `Spring` and `Decay` properties on `Animator`, and we should either **(a)** automatically switch between underlying `POPAnimation`s if necessary when a property is set; or **(b)** have an explicit `animationMode` property you use to "actuate" the switch. We can preserve velocity through the switch using the most recent `POPAnimationEvent`'s velocity.

What do you think? Too horrific? At the end of the day, I think this'll make things much nicer day-to-day.

saniul:
> Given that context, do the design decisions make more sense?
Ah, yes, thanks for clarifying!

>  I think this'll make things much nicer day-to-day.
Totally. 

> (a) automatically switch between underlying POPAnimations if necessary when a property is set;
*(a)* Sounds nicer, I guess.

I’m taking a break tomorrow (Wednesday), but I can work on this on Thursday, if it can wait till then.

andymatuschak:
There is no rush nor expectation of speed at all; thank you for volunteering!

andymatuschak:
Random thought: should we think of animations as a kind of behavior? After all, they apply every frame. And decay animations, in particular, are basically like a friction physics force. But something's making me uncertain there. Hm.

saniul:
Mmm, for some reason in my head behaviors (as in https://github.com/Khan/Prototope/pull/10) feel more reactionary, like “when Y happens, do X”.

Animations (or forces/physics constraints) on the other hand feel like actions – “do X [until Y]”.

Though I might be biased just because the first implementation of behaviors is the collision behavior.

andymatuschak:
To me behaviors are just “code associated with this layer that runs every frame.” Probably too vague though.

saniul:
Closing for now

PR #9, opened 2015-02-06
Particles!

jbrennan:
tumblr_inline_mk6dndpcx91qz4rgp

What it does
----------------

- Adds a wrapper around `CAEmitterCell` and `CAEmitterLayer` to provide layer-based particles.
- Adds presets for certain kinds of particles, because there’s a dizzying array of settings and it’s hard to know what’s going to look good.
- Adds a `Radian` type to make it more explicit when we’re dealing in radians. `Radian` is just a type alias for `Double`, but more semantically sound.

### DIFF
--- a/Prototope/Layer.swift
+++ b/Prototope/Layer.swift
@@ -387,6 +387,9 @@ public class Layer: Equatable, Hashable {
 	public func addParticleEmitter(particleEmitter: ParticleEmitter) {
 		self.particleEmitters.append(particleEmitter)
 		self.view.layer.addSublayer(particleEmitter.emitterLayer)
+		particleEmitter.emitterLayer.frame = self.view.layer.bounds
+		particleEmitter.size = self.size
+		particleEmitter.position = Point(particleEmitter.emitterLayer.position)
 		
 		// TODO(jb): Should we disable bounds clipping on self.view.layer or instruct devs to instead emit the particles from a parent layer?
 		self.view.layer.masksToBounds = false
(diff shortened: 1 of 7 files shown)


Prototope/Layer.swift
```diff
@@ -377,6 +377,41 @@ public class Layer: Equatable, Hashable {
...
+		self.view.layer.addSublayer(particleEmitter.emitterLayer)
+		particleEmitter.emitterLayer.frame = self.view.layer.bounds
+		particleEmitter.size = self.size
+		particleEmitter.position = Point(particleEmitter.emitterLayer.position)
+		
+		// TODO(jb): Should we disable bounds clipping on self.view.layer or instruct devs to instead emit the particles from a parent layer?
+		self.view.layer.masksToBounds = false
+		
+		if let duration = duration {
+			afterDuration(duration, { () -> Void in
```
andymatuschak: can be written as:

```
afterDuration(duration) {
  self.removeParticleEmitter(particleEmitter)
}
```

andymatuschak:
Looking good! Wish I could think of something nicer to do with the `foo` / `fooRange` APIs given Swift's newfound power, but nothing good comes to mind. Will noodle over lunch.

jbrennan:
@andymatuschak One option might be a “rangeable type” like:

    struct Rangeable<T> {
        let value: T
        let range: T
    }

But that might just be a pain in the butt.

Prototope/Layer.swift
jbrennan: Ah yes, autocomplete. I do love the “make it look like syntax” part.

andymatuschak:
Yeah, let's skip it for now.

PR #10, opened 2015-02-08
[FIN] Collision Behaviors

saniul:
...
---------
**UPDATE**: Changed API across the board with feedback from Andy.

**To do:**
- [x] Currently this only works for layers in the same parent layer. Need to make it work regardless of the layer hierarchy.
- [x] More feedback from Andy
- [x] Add JS bridging for the behavior API


Prototope/Collisions.swift
```diff
@@ -0,0 +1,184 @@
...
+        globalConstraintStore.registerConstraint(self)
+    }
+
+    func update() {
+        self.updateWithState(stateForLayer(layer1, andLayer: layer2))
+    }
+    
+    func updateWithState(state: ConstraintState) {
+        let old = previousState
+        switch (kind, old, state) {
```
andymatuschak: Dig this pattern.

Prototope/Collisions.swift
```diff
@@ -0,0 +1,184 @@
...
+    case Collides
+    
+    case IsContainedIn
+    case Contains
+    
+    case Enters
+    case Leaves
+}
+
+enum ConstraintState {
```
andymatuschak: Dig the explicit states sum type.

andymatuschak:
This is really cool, @saniul! Thank you for taking the initiative here.

I like the syntax, too, but given that I expect we're going to eventually add many features of this kind (i.e. some behavior which evaluates on every tick), I think we might want a little more structure to the interface so they don't all feel so different. What if we phrased "behaviors" like "gestures"?

```swift
layer.behaviors = [CollisionBehavior(on: .Entering(layer2)) { layer.sparkle() }]
```

Not quite as shiny DSL-y, I know. I think it might be best to play a little more conservative in that direction until we've lived on the API a bit longer.

---

Regarding value types vs. classes:

It does make some sense that you had to make `Constraint` be a reference type—it's not inert! In these kinds of situations, I find it's sometimes helpful to separate *the specification of the behaving entity* (i.e. the constraint configuration) from *the thing which behaves, using that specification* (i.e. the thing implementing `update()`.

saniul:
Thanks fortune feedback, I'll work on incorporating it tonight!

saniul:
s/fortune/for your

saniul:
How’s this?

Separated the thing into value types (`struct CollisionBehavior : protocol BehaviorType`) and reference types (`class CollisionBehaviorBinding : BehaviorBinding`). 

The value type is essentially a configuration structure for the binding. The binding contains all the state necessary to do whatever it needs to do.

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,194 @@
...
+//
+//  Created by Saniul Ahmed on 07/02/2015.
+//  Copyright (c) 2015 Khan Academy. All rights reserved.
+//
+
+import Foundation
+
+//MARK: Public Behavior Types
+
+// Protocol for types describing the "configuration" of a behavior.
```
andymatuschak: Mind using docstring syntax for these and the other comments here so they show up in Xcode autocomplete?

Prototope/Layer.swift
```diff
@@ -717,6 +717,12 @@ public class Layer: Equatable, Hashable {
 			wrappedLayer.frame = bounds
 		}
 	}
+    
+    public var behaviors: [BehaviorType] = [] {
+        didSet {
+            globalBehaviorStore.updateWithLayer(self, behaviors:behaviors)
```
andymatuschak: Nit: idiom is `behaviors: behaviors`

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,194 @@
...
+    func createBindingForLayer(layer: Layer, behavior: BehaviorType) -> BehaviorBinding? {
+        if let collisionBehavior = behavior as? CollisionBehavior {
+            return CollisionBehaviorBinding(hostLayer: layer, config: collisionBehavior)
+        }
+        
+        return nil
+    }
+}
+
+var globalBehaviorStore = BehaviorStore()
```
andymatuschak: With new live-reload features, we've gotta be more careful about globals. I know we've got a few more to clean up, but let's at least not add more. :)

To that end: would you mind moving this state to a field on the new `Environment` type? That way when we reload, the store will be cleared! :)

saniul: > With new live-reload features, we've gotta be more careful about globals. I know we've got a few more to clean up, but let's at least not add more. :)

Yup, this was implemented before I had a look at the shiny new stuff. 

I’ll do another pass on this PR to bring it up to date and incorporate any feedback :+1:

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,194 @@
...
+        self.otherLayer = otherLayer
+        self.handler = handler
+    }
+}
+
+//MARK: Behavior Bindings
+
+// Abstract. Describes a relationship between a layer and a single instance of a behavior.
+// Subclasses should encapsulate the state necessary to handle the behavior correctly.
+class BehaviorBinding: Equatable, Hashable {
```
andymatuschak: Could you make this a protocol (conforming to `Hashable`) instead of an abstract class? Those always feel kinda hacky. :)

Then `CollisionBehaviorBinding` could implement `==` with `===` rather than using a counter.

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,194 @@
...
+    
+    func fire() {
+        self.config.handler()
+    }
+}
+
+//MARK: Global Behavior Store
+
+// Global bag of layer behavior state
+class BehaviorStore {
```
andymatuschak: `Store` sounds like a value to me, but this thing behaves—it manages the heartbeat and owns control flow into the behaviors.

Perhaps `BehaviorDriver`?

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,194 @@
...
+    }
+    
+    func tick() {
+        for b in self.registeredBindings {
+            b.update()
+        }
+    }
+    
+    func updateWithLayer(layer: Layer, behaviors: [BehaviorType]) {
+        var knownBindings = lazy(self.registeredBindings).filter { $0.hostLayer == layer }
```
andymatuschak: s/var/let

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,194 @@
...
+        
+        let otherBindings = self.registeredBindings.subtract(knownBindings)
```
andymatuschak: Very nice.

andymatuschak:
Exciting!!

saniul:
:+1: Ok, I’ll finish this up tonight, instead of coming back to https://github.com/Khan/Prototope/pull/8. 

Thanks for feedback, @andymatuschak!

saniul:
Also, added "Add JS bridging” to the *To do*

andymatuschak:
No rush—we appreciate any contributions you can make!

Prototope/Behaviors.swift
saniul: Mmm, that was what I wanted to do initially, but:
`Equatable` uses `Self`, which makes the declaration `var registeredBindings: Set<BehaviorBinding>` in `BehaviorStore` impossible – “Protocol ‘BehaviorBinding’ can only be used as a generic constraint...”. This would force BehaviorStore to only work with a single type of behaviors, which is obviously not what we want.

If there’s a nicer way out of this other than an abstract class I would be very glad to learn it, this isn’t the first time I encounter this limitation.

saniul:
Alright, looks like this is it.

I also added a behavior which just runs on every frame:
```javascript
function makeBreathingLayer() {
	var spinnyLayer = gimmeSquare(324);
	var t = 0
	var behavior = new BlockBehavior({handler:function(layer) {
			var scale = 1+Math.cos(t)
			t = t+0.09
			layer.scale = scale
		}});
	spinnyLayer.behaviors = [behavior];
}
```
[Video of the above](http://cl.ly/1G2u1q3P3C3w).

 Nothing fancy - same thing could be achieved with a Heartbeat, but this lets us keep the behavior logic closer to the layers, reuse a single behavior across multiple layers, etc.

saniul:
@andymatuschak, just in case it got lost in the outdated diffs - I couldn't figure out a nice solution to https://github.com/Khan/Prototope/pull/10#discussion_r24544435

PrototopeJSBridge/BehaviorBridge.swift
```diff
@@ -0,0 +1,142 @@
...
+//
+//  Created by Saniul Ahmed on 11/02/2015.
+//  Copyright (c) 2015 Khan Academy. All rights reserved.
+//
+
+import Foundation
+import JavaScriptCore
+import Prototope
+
+func behaviorBridgeForBehavior(behavior: BehaviorType) -> BehaviorBridgeType {
```
andymatuschak: So wish we didn't need these!

saniul: I know :(

saniul:
Rebased on master to resolve the merge conflict

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,218 @@
...
+import Foundation
+
+//MARK: Public Behavior Types
+
+/** Protocol for types describing the "configuration" of a behavior. */
+public protocol BehaviorType {}
+
+/** Describes a behavior that just calls a closure/block on every heartbeat,
+    passing the host layer to the closure. */
+public struct BlockBehavior: BehaviorType {
```
andymatuschak: Got some feedback from the designers: since "blocks" aren't a thing in JS, this naming was a bit surprising. How about `ActionBehavior`?

Prototope/Behaviors.swift
```diff
@@ -0,0 +1,218 @@
...
+    public enum Kind {
+        case Entering
+        case Leaving
+    }
+    
+    let kind: Kind
+    let otherLayer: Layer
+    let handler: () -> Void
+    
+    public init(on kind: Kind, _ otherLayer: Layer, handler: ()->Void) {
```
andymatuschak: Got some feedback from the designers. They'd actually prefer that `CollisionBehavior` calls `handler` on all `kind`s of events but passes `kind` in—they said they imagined they'd have to create both an entering and a leaving behavior most of the time. What do you think?

Prototope/Environment.swift
```diff
@@ -12,6 +12,7 @@ import Foundation
 public struct Environment {
 	public var rootLayer: Layer
 	public var imageProvider: String -> UIImage?
+    var behaviorDriver: BehaviorDriver
```
andymatuschak: It's so nice and simple!

PrototopeJSBridge/BehaviorBridge.swift
```diff
@@ -0,0 +1,142 @@
...
+    
+    public init(_ behavior: CollisionBehavior) {
+        self.collisionBehavior = behavior
+        super.init()
+    }
+    
+    required public init?(args: JSValue) {
+        let kindValue = args.valueForProperty("kind")
+        let otherLayerValue = args.valueForProperty("otherLayer")
+        let handler = args.objectForKeyedSubscript("handler")
```
andymatuschak: This actually needs to be `args.valueForProperty("handler")` because you need to leave it as a `JSValue` to `callWithArguments` on line 61.

PrototopeJSBridge/BehaviorBridge.swift
```diff
@@ -0,0 +1,142 @@
...
+        
+        if let kind = CollisionBehaviorKindBridge.decodeKind(kindValue) where !kindValue.isUndefined(),
+            let otherLayer = (otherLayerValue as? JSExport as? LayerBridge)?.layer where !otherLayerValue.isUndefined(),
```
andymatuschak: Here `otherLayerValue` is a `JSValue`, so this cast will always fail. I had to write `otherLayerValue.toObject()` to get to the `LayerBridge` here:

```
-            let otherLayer = (otherLayerValue as? JSExport as? LayerBridge)?.layer where !otherLayerValue.isUndefined(),
+            let otherLayer = (otherLayerValue.toObject() as! JSExport as! LayerBridge).layer where !otherLayerValue.isUndefined(),
```

PrototopeJSBridge/BehaviorBridge.swift
```diff
@@ -0,0 +1,142 @@
...
+    enum RawKind: Int {
+        case Entering = 0
+        case Leaving
+    }
+    
+    public class func addToContext(context: JSContext) {
+        let kindObject = JSValue(newObjectInContext: context)
+        kindObject.setObject(RawKind.Entering.rawValue, forKeyedSubscript: "Entering")
+        kindObject.setObject(RawKind.Leaving.rawValue, forKeyedSubscript: "Leaving")
+        context.setObject(kindObject, forKeyedSubscript: "Kind")
```
andymatuschak: Can we call this `CollisionKind`? Or maybe it should be `CollisionBehavior.Kind`? Having `Kind` at the root feels weird.

PrototopeJSBridge/Context.swift
```diff
@@ -72,5 +72,7 @@ public class Context {
 		ParticleBridge.addToContext(context)
 		ParticleEmitterBridge.addToContext(context)
 		ScrollLayerBridge.addToContext(context)
+        CollisionBehaviorBridge.addToContext(context)
+        BlockBehaviorBridge.addToContext(context)
```
andymatuschak: Gotta add `CollisionBehaviorKindBridge` too!

andymatuschak:
Very exciting! I also modified the Behaviors example to have a collision behavior, in case you'd like to test the bridge:

```
+var breathingLayer = makeBreathingLayer()
+
+var square = gimmeSquare(75)
+square.gestures = [
+       new PanGesture({
+               handler: function(phase, sequence) {
+                       if (sequence.previousSample !== undefined) {
+                               var current = sequence.currentSample.globalLocation
+                               var previous = sequence.previousSample.globalLocation
+                               square.position = square.position.add(current.subtract(previous))
+                       }
+               }
+       })
+]
+
+square.behaviors = [
+       new CollisionBehavior({
+               kind: Kind.Entering,
+               otherLayer: breathingLayer,
+               handler: function() {
+                       square.animators.backgroundColor.target = Color.yellow
+               }
+       }),
+       new CollisionBehavior({
+               kind: Kind.Leaving,
+               otherLayer: breathingLayer,
+               handler: function() {
+                       square.animators.backgroundColor.target = Color.white
+               }
+       })
+
+]
```

Prototope/Behaviors.swift
saniul: Sounds good!

PrototopeJSBridge/BehaviorBridge.swift
saniul: Argh, didn’t have time to write tests last night. I know I should’ve!

PrototopeJSBridge/BehaviorBridge.swift
saniul: Yes, makes sense.

saniul: Opted for CollisionBehaviorKind

saniul:
Done!

Also incorporated your main.js changes:
```javascript
square.behaviors = [
       new CollisionBehavior({
               with: breathingLayer,
               handler: function(kind) {
					if (kind == CollisionBehaviorKind.Entering) {
                       square.animators.backgroundColor.target = Color.yellow
					} else if (kind == CollisionBehaviorKind.Leaving) {
						square.animators.backgroundColor.target = Color.white
					}
               }
       }),
]
```

andymatuschak:
Very nicely done. :)

saniul:
:boom: Woot!

PR #14, opened 2015-02-11
Particle bridge

jbrennan:
Bridges particles and Particle Emitter

## Todo

- Figure out why the layer bridge’s particle methods aren’t…well, bridged.

- Eventually add particle presets.

### DIFF
--- a/PrototopeJSBridge/LayerBridge.swift
+++ b/PrototopeJSBridge/LayerBridge.swift
@@ -76,6 +76,10 @@ import JavaScriptCore
 
     // MARK: Animations
     var animators: LayerAnimatorStoreJSExport { get }
+	
+	// MARK: Particle emitters
+	func addParticleEmitter(emitterBridge: ParticleEmitterBridge)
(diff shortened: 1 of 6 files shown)


PrototopeJSBridge/LayerBridge.swift
```diff
@@ -76,6 +76,10 @@ import JavaScriptCore
 
     // MARK: Animations
     var animators: LayerAnimatorStoreJSExport { get }
+	
+	// MARK: Particle emitters
+	func addParticleEmitter(emitterBridge: ParticleEmitterBridge)
```
andymatuschak: Unfortunately the reason these aren't bridging is that you have to type these as `emitterBridge: ParticleEmitterJSExport`, not `ParticleEmitterBridge`. This… took me a while to figure out.

jbrennan: Oh! Thank you, I’ll fix those up then.

PR #15, opened 2015-02-11
[Hold merge] Scroll layer

jbrennan:
Adds a scrolly layer!


andymatuschak:
There's something super-weird going on with the merge commits there, but this is good to go! We'll probably need to extend this over time, but it's definitely a good enough start for now. Between this and #8 we're gonna be in good inertial shape.

PR #16, opened 2015-02-13
[FIN] Replace “siblingView” with “siblingLayer" in the Layout extension

saniul:


andymatuschak:
Whoooops. Good call!

/cc @jbrennan

jbrennan:
Whoops, missed one! Thanks!

PR #18, opened 2015-02-16
Adding text layers

andymatuschak:
Here’s an API. What do you folks think?

- [ ] add ghpages docs

gif of text layer scene

/cc @jbrennan @may-li @nefaurk @nsfmc @saniul

### DIFF
--- a/PrototopeJSBridge/VideoLayerBridge.swift
+++ b/PrototopeJSBridge/VideoLayerBridge.swift
@@ -11,26 +11,25 @@ import Prototope
 import JavaScriptCore
 
 @objc public protocol VideoLayerJSExport: JSExport {
-	init(args: NSDictionary)
+	init?(args: NSDictionary)
 	func play()
 	func pause()
 }
 
 
 @objc public class VideoLayerBridge: LayerBridge, VideoLayerJSExport, BridgeType {
-	var videoLayer: VideoLayer
+	var videoLayer: VideoLayer { return layer as! VideoLayer }
 	
 	public override class func addToContext(context: JSContext) {
 		context.setObject(self, forKeyedSubscript: "VideoLayer")
 	}
 	
-	required public init(args: NSDictionary) {
+	required public init?(args: NSDictionary) {
 		let parentLayer = (args["parent"] as! LayerBridge?)?.layer
 		let video = (args["video"] as! VideoBridge?)?.video
 		
-		self.videoLayer = VideoLayer(parent: parentLayer, video: video)
-
-		super.init(args: args)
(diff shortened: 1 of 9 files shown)


saniul:
Yep, this is pretty much how I imagined this.

saniul:
Also, this calls for a text-alignment property.

saniul:
Suggestion, what if we always set `attributedText` under the hood, instead of just `text`? We can also expose an HTML-text setter and use `[[NSAttributedString alloc] initWithData...` internally. Designers can them provide styled strings using that method.

saniul:
https://github.com/alexfish/stylize is also relevant

saniul:
> Also, this calls for a text-alignment property.

Implemented in https://github.com/Khan/Prototope/pull/19

andymatuschak:
Wooo, thanks for making the text alignment pull request! Will take a look at that shortly.

I'd like to wait on exposing fancier styling until we have a better idea of how we'd use it. Would prefer something more structured than HTML (Stylize is a nice reference), but maybe HTML's the easiest/familiar thing. @may-li, @nefaurk, what do you think?

PrototopeJSBridge/VideoLayerBridge.swift
```diff
@@ -11,26 +11,25 @@ import Prototope
...
-	required public init(args: NSDictionary) {
+	required public init?(args: NSDictionary) {
 		let parentLayer = (args["parent"] as! LayerBridge?)?.layer
 		let video = (args["video"] as! VideoBridge?)?.video
 		
-		self.videoLayer = VideoLayer(parent: parentLayer, video: video)
-
-		super.init(args: args)
+		let videoLayer = VideoLayer(parent: parentLayer, video: video)
+		super.init(videoLayer)
```
andymatuschak: @jbrennan Here and in ScrollLayerBridge, I made this fix: the previous code actually ended up initializing two layers—the video layer and also the layer in super's `init` implementation.

Implementation inheritance. *sigh*

andymatuschak:
Oh, uh, I totally forgot the ghpages docs. Will do that shortly.

PrototopeJSBridge/VideoLayerBridge.swift
jbrennan: Oh deuce.

nsfmc:
my only comment, mostly for docs, is that there should be an easy way to know where the 1st baseline goes for a given textlayer (i.e. is it the layer's y minus the default leading plus the descender?)

saniul:
@nsfmc I'll do this tonight when I get home

saniul:
^ https://github.com/Khan/Prototope/pull/24

PR #20, opened 2015-02-17
[FIN] Make `AnimationDelegate` available to pop

saniul:
pop_animationDidStop was never called by the pop framework

### DIFF
--- a/Prototope/Animation.swift
+++ b/Prototope/Animation.swift
@@ -212,10 +212,10 @@ extension Layer {
 
 // MARK: - Internal interfaces
 
-@objc private class AnimationDelegate: NSObject, POPAnimationDelegate {
+@objc public class AnimationDelegate: NSObject, POPAnimationDelegate {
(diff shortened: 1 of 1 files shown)


Prototope/Animation.swift
```diff
@@ -212,10 +212,10 @@ extension Layer {
 
 // MARK: - Internal interfaces
 
-@objc private class AnimationDelegate: NSObject, POPAnimationDelegate {
+@objc public class AnimationDelegate: NSObject, POPAnimationDelegate {
 	var completionHandler: (() -> Void)?
 
-	func pop_animationDidStop(animation: POPAnimation, finished: Bool) {
+	public func pop_animationDidStop(animation: POPAnimation, finished: Bool) {
```
andymatuschak: Oooooh, I got bit by so many times! You should be able to keep the class private and make this work by declaring the function as `@objc func pop_animationDidStop`

jbrennan:
Thank you. Will merge shortly.

PR #21, opened 2015-02-17
[WIP] Emoji layers

saniul:
Provides a simpler interface than text layer.

Jumping emojis

To do:
- [ ] Allow modifying the set emoji
- [ ] Test the bridge
- [ ] Document `EmojiLayer`s interface


NachoSoto:
LOL :D

jbrennan:
I’m curious, how is this different than using a text layer with an emoji string?

saniul:
@jbrennan: Essentially no difference, except limiting the public interface to things that are only relevant when dealing with single `Character`s, as opposed to multi-`Character` `String`s.

I was thinking about adding an additional behavior, which *mostly* makes sense for this class and not for `TextLayer`: say, you change the frame of the `EmojiLayer` to a given value, then call `updateFontSizeAutomatically`, and it selects the largest font size that fits the current frame.

andymatuschak:
Hm, while I appreciate the contribution, I'm not yet sure this is something we'll use enough to justify the maintenance burden. What do you think about waiting to see how much this comes up?

saniul:
Of course. I’ll close the PR for now.

PR #22, opened 2015-02-17
Speech

jbrennan:
Now your prototopes can really say something!

### DIFF
--- a/PrototopeJSBridge/SpeechBridge.swift
+++ b/PrototopeJSBridge/SpeechBridge.swift
@@ -1,0 +1,40 @@
+//
+//  SpeechBridge.swift
+//  Prototope
+//
+//  Created by Jason Brennan on 2015-02-17.
+//  Copyright (c) 2015 Khan Academy. All rights reserved.
+//
+
+import Foundation
+import Prototope
+import JavaScriptCore
+
+@objc public protocol SpeechJSExport: JSExport {
+}
+
+@objc public class SpeechBridge: NSObject, SpeechJSExport, BridgeType {
+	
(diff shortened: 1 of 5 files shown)


PrototopeJSBridge/SpeechBridge.swift
```diff
@@ -0,0 +1,40 @@
...
+		context.setObject(self, forKeyedSubscript: "Speech")
+		let speechBridge = context.objectForKeyedSubscript("Speech")
+		
+		speechBridge.setFunctionForKey("say", fn: sayTrampoline)
+	}
+	
+}
+
+
+let sayTrampoline: @objc_block JSValue -> Void = { args in
```
NachoSoto: Whoa, `@objc_block`? TIL!

saniul: @NachoSoto Also worth mentioning – https://github.com/Khan/Prototope/pull/20#discussion_r24793735

jbrennan: Yeah, TIL too! We use it elsewhere in prototope for animation bridging I think.

NachoSoto:
Not a fan of types with all static methods, but it's probably the most convenient API for this :)

jbrennan:
@NachoSoto Yeah, I went back and forth on that. I settled on static methods because it really only makes sense to speak one thing at a time, and this lets you queue things up without having to figure out when to make the next thing speak.

NachoSoto:
Yup! :shipit: 
Should I merge? since you assigned it to me

jbrennan:
@NachoSoto Yep!

PR #19, opened 2015-02-16
Text alignment for TextLayers

saniul:
This is a companion PR to https://github.com/Khan/Prototope/pull/18

PR #23, opened 2015-02-18
[FIN] TextLayer – Alignment

saniul:
https://github.com/Khan/Prototope/pull/19 was only merged into `text-layers` but never made it to `master`


saniul:
This is ready to go

PR #24, opened 2015-02-18
[FIN] TextLayer – Baseline Enhancements

saniul:
Depends on https://github.com/Khan/Prototope/pull/23 (Well, doesn’t actually depend, but contains commits from that – all the unrelated TextAlignment stuff).

* getter for the offset from the top of a `TextLayer` to its first line’s baseline.
* convenience function to align a `TextLayer` to another `TextLayer`’s first line’s baseline.
* device-pixel-aware `ceil` and `floor` functions

### DIFF
--- a/Examples/TestText/main.js
+++ b/Examples/TestText/main.js
@@ -20,6 +20,14 @@ wrappingLayer.x = Layer.root.x
 wrappingLayer.y = 300
 wrappingLayer.width = 100
 
+var alignmentLayer = new TextLayer()
+alignmentLayer.text = "align"
+alignmentLayer.fontSize = 64
+alignmentLayer.border = new Border({color: Color.red, width: 1})
+
 var h = new Heartbeat({handler: function(heartbeat) {
 	wrappingLayer.width = 100 + Math.sin(heartbeat.timestamp * 4) * 50
+	alignmentLayer.alignWithBaselineOf(wrappingLayer)
+	alignmentLayer.x = wrappingLayer.x + wrappingLayer.width*0.5 + alignmentLayer.width*0.5
+
 }})
--- a/Prototope/TextLayer.swift
+++ b/Prototope/TextLayer.swift
@@ -111,7 +111,7 @@ public class TextLayer: Layer {
 		return Double(label.font.ascender)
 	}
 	
-	/** Aligns this layer with the first baseline of the other layer */
+	/** Aligns this layer's first baseline with the first baseline of the other layer */
 	public func alignWithBaselineOf(otherLayer: TextLayer) {
 		var delta = pixelAwareCeil(otherLayer.baselineHeight-baselineHeight)
 		
--- a/Prototope/TextLayer.swift
+++ b/Prototope/TextLayer.swift
@@ -105,7 +105,21 @@ public class TextLayer: Layer {
 			//No need to adjust size, since changing alignment doesn't influence it
 		}
 	}
-
+	
+	/** Distance from top of layer to the first line's baseline */
+	public var baselineHeight: Double {
+		return Double(label.font.ascender)
+	}
+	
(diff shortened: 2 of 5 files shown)


saniul:
@nsfmc, is this what you envisioned?
```swift
layerB.alignWithBaselineOf(layerA)
```

If so, then this is ready for review.

Prototope/TextLayer.swift
```diff
@@ -56,7 +95,31 @@ public class TextLayer: Layer {
...
+		return Double(label.font.ascender)
+	}
+	
+	/** Aligns this layer with the first baseline of the other layer */
+	public func alignWithBaselineOf(otherLayer: TextLayer) {
+		var delta = pixelAwareCeil(otherLayer.baselineHeight-baselineHeight)
+		
+		var frame = self.frame
+		frame.origin.y = otherLayer.frame.minY + delta
+		self.frame = frame
```
NachoSoto: In Swift you can actually change the property you want:
self.frame.origin.y = ...

Prototope/TextLayer.swift
```diff
@@ -56,7 +95,31 @@ public class TextLayer: Layer {
...
+		}
+	}
+	
+	/** Distance from top of layer to the first line's baseline */
+	public var baselineHeight: Double {
+		return Double(label.font.ascender)
+	}
+	
+	/** Aligns this layer with the first baseline of the other layer */
+	public func alignWithBaselineOf(otherLayer: TextLayer) {
```
NachoSoto: Seems weird to have a one-off layout method for this Layer. Not knowing what the use case of baseline will be, I'd say we can leave this out for now. What do you think?

Prototope/TextLayer.swift
saniul: Ah yes, not sure what was happening here. the `delta` should also be a constant, not a var (I blame myself for copypasting this from the js scene, where I implemented this logic initially).

Will fix in a sec.

Prototope/TextLayer.swift
saniul: Hm, this seems like a very common use-case (if not *the* most common use case for using the baseline). My guess was that @nsfmc was asking for a baseline getter (https://github.com/Khan/Prototope/pull/18#issuecomment-74729063) to do precisely this, so I’d be interested in hearing his feedback?

Also, it might make sense to rename this method to be more consistent with https://github.com/Khan/Prototope/blob/7f06c827c10fc36f233a3d791485c17644fd2331/Prototope/Layer%2BLayout.swift

Examples/TestText/main.js
```diff
@@ -19,6 +20,14 @@ wrappingLayer.x = Layer.root.x
...
 
+var alignmentLayer = new TextLayer()
+alignmentLayer.text = "align"
+alignmentLayer.fontSize = 64
+alignmentLayer.border = new Border({color: Color.red, width: 1})
+
 var h = new Heartbeat({handler: function(heartbeat) {
 	wrappingLayer.width = 100 + Math.sin(heartbeat.timestamp * 4) * 50
+	alignmentLayer.alignWithBaselineOf(wrappingLayer)
+	alignmentLayer.x = wrappingLayer.x + wrappingLayer.width*0.5 + alignmentLayer.width*0.5
```
jbrennan: We also now have methods like `moveToRightOfSiblingLayer()` if that’s more useful than layout via centre/frame math :)

saniul: I think those functions aren't bridged to JS, that's next on my PR to do list :)

jbrennan: Oh right, I didn’t realize this was JS. Don’t worry about bridging them, I’m going to jump in and do that one. If you could fix up the merge conflicts, this is good to go!

saniul:
Rebased on master, cleared up the merge conflict

PR #27, opened 2015-02-19
[FIN] Layout Bridge Bugfix + Additional Layout Methods

saniul:
* Fixed LayerBridge.frameMaxY
* Added missing `moveToXXXSideOfParentLayer` methods

### DIFF
--- a/Prototope/Layer+Layout.swift
+++ b/Prototope/Layer+Layout.swift
@@ -69,7 +69,7 @@ extension Layer {
 	/** Moves the receiver so that its left side is aligned with the left side of its parent layer. */
 	public func moveToLeftSideOfParentLayer(margin: Double = 0.0) {
 		if let parent = self.parent {
-			self.frameMinX = -margin
+			self.frameMinX = margin
 		}
 	}
 	
@@ -77,7 +77,7 @@ extension Layer {
 	/** Moves the receiver so that its top side is aligned with the top side of its parent layer. */
 	public func moveToTopSideOfParentLayer(margin: Double = 0.0) {
 		if let parent = self.parent {
-			self.frameMinY = -margin
+			self.frameMinY = margin
 		}
 	}
 	
--- a/Prototope/Layer+Layout.swift
+++ b/Prototope/Layer+Layout.swift
@@ -69,7 +69,7 @@ extension Layer {
 	/** Moves the receiver so that its left side is aligned with the left side of its parent layer. */
 	public func moveToLeftSideOfParentLayer(margin: Double = 0.0) {
 		if let parent = self.parent {
-			self.frameMinX = 0 + margin
+			self.frameMinX = -margin
 		}
 	}
 	
@@ -77,7 +77,7 @@ extension Layer {
 	/** Moves the receiver so that its top side is aligned with the top side of its parent layer. */
 	public func moveToTopSideOfParentLayer(margin: Double = 0.0) {
 		if let parent = self.parent {
-			self.frameMinY = 0 + margin
+			self.frameMinY = -margin
 		}
 	}
 	
--- a/Prototope/Layer+Layout.swift
+++ b/Prototope/Layer+Layout.swift
@@ -66,6 +66,30 @@ extension Layer {
 	}
 	
 	
+	/** Moves the receiver so that its left side is aligned with the left side of its parent layer. */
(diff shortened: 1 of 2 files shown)


saniul:
@jbrennan

Prototope/Layer+Layout.swift
```diff
@@ -66,6 +66,30 @@ extension Layer {
 	}
 	
 	
+	/** Moves the receiver so that its left side is aligned with the left side of its parent layer. */
+	public func moveToLeftSideOfParentLayer(margin: Double = 0.0) {
+		if let parent = self.parent {
+			self.frameMinX = 0 + margin
```
jbrennan: Any particular reasons for having `0 + margin`?

PrototopeJSBridge/LayerBridge.swift
```diff
@@ -577,7 +580,7 @@ import JavaScriptCore
 	
 	
 	/** The maxY of the layer's frame. */
-	public var frameMaxY: Double { return self.layer.frameMinY }
```
jbrennan: Doh.

Prototope/Layer+Layout.swift
saniul: D'oh, that should be -margin, to match the other methods. Fixed.

Prototope/Layer+Layout.swift
```diff
@@ -66,6 +66,30 @@ extension Layer {
 	}
 	
 	
+	/** Moves the receiver so that its left side is aligned with the left side of its parent layer. */
+	public func moveToLeftSideOfParentLayer(margin: Double = 0.0) {
+		if let parent = self.parent {
+			self.frameMinX = -margin
```
jbrennan: I think this is backwards still. This method tells me “Move my originX (frameMinX) to the leftmost side of my parent (i.e., x = 0) but then offset it with a margin towards the right.

```
layer.moveToLeftSideOfParentLayer() // aligns the layer flush against its parent’s originX
layer.moveToLeftSideOfParentLayer(margin: 10) // essentially sets the layer’s originX to be 10
```

Prototope/Layer+Layout.swift
```diff
@@ -66,6 +66,30 @@ extension Layer {
...
+		}
+	}
+	
+	
+	/** Moves the receiver so that its top side is aligned with the top side of its parent layer. */
+	public func moveToTopSideOfParentLayer(margin: Double = 0.0) {
+		if let parent = self.parent {
+			self.frameMinY = -margin
```
jbrennan: Same goes for this one.

Prototope/Layer+Layout.swift
saniul: I was made it so it matches with `moveToLeftOfSiblingLayer`:
```swift
	public func moveToLeftOfSiblingLayer(siblingLayer: Layer, margin: Double = 0.0) {
		self.frameMinX = floor(siblingLayer.frameMinX - (self.width + margin))
	}
```

saniul: Same with `moveToTopSideOfParentLayer` and `moveAboveSiblingLayer`.

saniul: I don’t mind changing it though, just trying to keep related things consistent :)

jbrennan: Aha. `moveToLeftOfSiblingLayer` is different because it’s with respect to the given `siblingLayer`, whereas `moveToLeftSideOfParentLayer` moves with respect to the `parent` layer.

If you have a layer called `parent`, with child layers `child1` and `child2`, `child1.moveToLeftSideOfParentLayer()` will move to be aligned with parent layer’s origin (you can think of this as “stick child to the left side of its parent” where `child1.moveToLeftOfSiblingLayer(child2)` will place `child1` beside `child2` on the left side.

I’ll grant it’s a bit confusing at first glance and I’m probably blinded by being too used to them :smile: but they’re pretty nice once you get the swing of them.

saniul: Right, yeah, I understood the logic for siblings but it only just now clicked for me that it makes sense for the margin in `parent` methods to mean “margin towards the centre of the layer”.

Made the change.

jbrennan: Yahoo! Thanks so much, and sorry for the confusion :)

Issue #28, opened 2015-02-25
Protoscope sometimes can’t find images


jbrennan:
I’m not sure why this is, but it ends up giving us a fatalError in the Image type because it can’t always load the image (even though it’s present).

andymatuschak:
Yikes. Can you post an example that exhibits the issue?

jbrennan:
Hmmm. OK I think I figured it out. The problem is when you broadcast just a `js` file, not the directory *containing* the `js` file.

andymatuschak:
Oh, that makes sense. I basically didn't intend that to work, but now that you point it out, maybe it should.

My thinking was that if you drop a .js on Protocaster, it just broadcasts that file. Maybe it should broadcast everything in the folder? Is that weird?

At minimum, we should propagate these kinds of errors through `Environment.exceptionHandler` so that e.g. Protoscope can display the error as a protonope and send it back to the host.

saniul:
> Maybe it should broadcast everything in the folder? Is that weird?

I think it’s totally reasonable to load the whole thing when only main.js was selected. In fact, it’s what I expected it to do the first time I tried it and encountered the same problem @jbrennan did.

andymatuschak:
OK!

PR #31, opened 2015-02-28
[FIN] Smarter scene URL selection

saniul:
If a file is selected, then the containing directory will be treated as selected. This should prevent situations where a user will accidentally select a main.js file instead of the directory which contains the rest of the scene's resources.

### DIFF
--- a/Protocaster/AppDelegate.swift
+++ b/Protocaster/AppDelegate.swift
@@ -32,9 +32,22 @@ class AppDelegate: NSObject, NSApplicationDelegate, DTBonjourDataConnectionDeleg
 			}
 		}
 	}
+    
+    private final func sceneURL(fromURL URL: NSURL) -> NSURL? {
+        var isDirectoryValue: AnyObject?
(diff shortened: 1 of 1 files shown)


saniul:
This resolves https://github.com/Khan/Prototope/issues/28

Protocaster/AppDelegate.swift
```diff
@@ -32,9 +32,22 @@ class AppDelegate: NSObject, NSApplicationDelegate, DTBonjourDataConnectionDeleg
...
+        if !isDirectory {
+            return URL.URLByDeletingLastPathComponent
+        }
+        
+        return URL
+    }
 
 	private func selectedPathDidChange(newURL: NSURL?) {
-		if let URL = newURL {
+		if let URL = (newURL >>- sceneURL) {
```
andymatuschak: Can we start phasing out usage of `bind` in favor of e.g. Swift 1.2's multiple destructuring if-let?

andymatuschak:
Looks good!

saniul:
@andymatuschak, ready to merge

andymatuschak:
Thanks!

PR #36, opened 2015-03-03
JS prototypes use JS strict mode

andymatuschak:
Tested all our JS prototypes to make sure they still function as expected. Fixed one; the rest look good.

This will make us throw protonopes for common errors (e.g. referring to variables that don't exist).

### DIFF
--- a/Examples/Color Puddles/main.js
+++ b/Examples/Color Puddles/main.js
@@ -39,7 +39,7 @@ touchLayer.touchEndedHandler = touchLayer.touchCancelledHandler = function(touch
 }
 
 var h = new Heartbeat({handler: function() {
-	for (layer in touchLayers) {
+	for (var layer in touchLayers) {
 		touchLayers[layer].scale *= 1.11
 	}
 }})
--- a/PrototopeJSBridge/Context.swift
+++ b/PrototopeJSBridge/Context.swift
@@ -31,7 +31,7 @@ public class Context {
 	}
 
 	public func evaluateScript(script: String!) -> JSValue {
-		return context.evaluateScript(script)
+		return context.evaluateScript("\"use strict\";" + script)
 	}
 
 	private func addBridgedTypes() {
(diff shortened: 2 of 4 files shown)


Examples/Color Puddles/main.js
```diff
@@ -39,7 +39,7 @@ touchLayer.touchEndedHandler = touchLayer.touchCancelledHandler = function(touch
 }
 
 var h = new Heartbeat({handler: function() {
-	for (layer in touchLayers) {
+	for (var layer in touchLayers) {
```
saniul: Any reason for this `var` here?

saniul: Darn it, this is JavaScript, not swift. Nevermind then.

NachoSoto: `for (layer in touchLayers)` would put `layer` into the global context: `window`. `var` ensures that it's only declared locally.

*Sort of off-topic:*
Although `var` "suffers" from hoisting. `let` is actually preferred since ES6.

PrototopeJSBridge/Context.swift
```diff
@@ -31,7 +31,7 @@ public class Context {
 	}
 
 	public func evaluateScript(script: String!) -> JSValue {
-		return context.evaluateScript(script)
+		return context.evaluateScript("\"use strict\";" + script)
```
NachoSoto: :boom:

PrototopeJSBridgeTests/ContextTests.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+import PrototopeJSBridge
+import XCTest
+
+class ContextTests: JSBridgeTestCase {
+	func testContextExecutesInStrictMode() {
+		let context = Context()
+		var hitException = false
+		context.exceptionHandler = { _ in hitException = true }
+		context.evaluateScript("x = 4")
```
NachoSoto: Clever way to test it :D

PR #39, opened 2015-03-03
The root layer is now *hosted* by a root view, but does not wrap that view.

### LINKED ISSUE #37: Root layer touchesHandlers don't get called
You can't install touch handlers (e.g. `touchBeganHandler`) on the root layer of a prototype because it's a "captured" non-Prototope-controlled view, which means its view isn't a `TouchForwardingImageView`. We'll need to add an extra layer of indirection

jbrennan:
This is done so that the root layer in the layer tree works exactly like every other node in the tree. It is no longer special (i.e., broken) because its view is no longer at the top of the view hierarchy.

[Fixes #37]

### DIFF
--- a/Prototope/Environment.swift
+++ b/Prototope/Environment.swift
@@ -18,7 +18,7 @@ public struct Environment {
 	public static var currentEnvironment: Environment?
 
 	public init(rootView: UIView, imageProvider: String -> UIImage?, soundProvider: String -> NSData?) {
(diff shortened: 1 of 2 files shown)


andymatuschak:
!! :heart:

jbrennan:
Feel free to delete branches after PRs get merged :D

andymatuschak:
ooh yeah

PR #45, opened 2015-03-06
[FIN] Be more graceful when encountering non-fatal errors

saniul:


Prototope/Image.swift
```diff
@@ -20,12 +20,12 @@ public struct Image {
...
 
 	/** Loads a named image from the assets built into the app. */
-	public init!(name: String) {
+	public init?(name: String) {
 		if let image = Environment.currentEnvironment!.imageProvider(name) {
 			uiImage = image
 			self.name = name
 		} else {
-			fatalError("Image named \(name) not found")
+			Environment.currentEnvironment?.exceptionHandler("⚠️ Image named \(name) not found")
```
andymatuschak: I love the use of the ⚠️ prefix, but I'd suggest it might better belong in the default implementation of the exception handler (which logs to the console) at `Environment.swift:55`, rather than being included in every exception string.

Then Protoscope (which already has formatting and special display) for exceptions can avoid redundancy, etc.

andymatuschak:
Most excellent!

saniul:
> I love the use of the ⚠️ prefix, but I'd suggest it might better belong in the default implementation of the exception handler

Done!

andymatuschak:
\o/

Issue #40, opened 2015-03-05
TextLayer.textAlignment should affect the way the text layer grows


andymatuschak:
When you change the `text` value of a `TextLayer`, it grows to accommodate, but right now it always grows so as to keep the same left edge. I'd expect (and @may-li just ran into this) that if you set the alignment to `Center`, it would grow by maintaining its center position, and so on.

PR #46, opened 2015-03-06
[FIN] TextLayer.textAlignment should affect the way the text layer grows

saniul:
Resolves https://github.com/Khan/Prototope/issues/40


andymatuschak:
Great work!

Issue #25, opened 2015-02-18
Prototope, Xcode, and Protocaster all crash when you forget “new” in your script


jbrennan:
This one was fun. I had the following line in my script,

```
var layer = TextLayer()
```

And I couldn’t figure out what was wrong with it. Turns out I forgot `new`. Anyway, when Prototope hit that line, it crashed, and sometimes this caused Xcode to crash (betas!). Obviously the script is wrong but for some reason Prototope didn’t catch the exception...not quite sure why yet.

saniul:
Investigated a bit – looks like the constructor functions that get created for the `JSExport`ed classes are generated in such a weird way that they crash JavaScriptCore when called straight, without the `new` keyword. If you look at the `JSExport.h` documentation, it actually states that those constructors aren’t 
callable:

![screen shot 2015-03-06 at 00 16 47](https://cloud.githubusercontent.com/assets/371773/6522154/21abb4a2-c396-11e4-9e5a-4bbaa4702125.png)

This is an actual crash (not a thrown exception) in JavaScriptCore, couldn’t figure out a workaround :( We can either live with it or stop using the constructor and expose a `create` method on all bridged types. I think just making sure we always write `new` is a better option, despite the crashiness.

andymatuschak:
Yikes. I've got no bright ideas here, other than e.g. preprocessing the JS to look for this kind of mistake, which is probably overboard. @jbrennan?

kevinb7:
It's probably not a bad idea to do some sort of linting before running.  If we're looking to support JavaScript as a language it shouldn't look too different from regular JavaScript that you see in the browser/node.  For instance, I was surprised that there was no ```setTimeout```, but instead there's a ```afterDuration``` function.  Hopefully we can support ```new``` as well as more of the JavaScript standard library in the future.

NachoSoto:
How do we define the constructors? I don't know if this is possible because I haven't really looked at the bridging code, but you can detect in a constructor if it's been invoked correctly by checking `if (this instanceof Constructor)`

saniul:
Constructor objects are defined by JavaScriptCore automatically based on the exposed `init` in the XXXJSBridge protocol (the one that extends JSExport).

spicyj:
Just seeing this in passing via email so I haven't looked at how the code is set up, but if you can run something along these lines before the user code, perhaps it will help:

```
function wrapNativeClass(OriginalConstructor) {
  function Constructor() {
[...]

PR #47, opened 2015-03-07
POC: Wrap bridged types to throw when called as a function

spicyj:
For #25: This shows how we might wrap the constructors so that they throw when called without `new`. In this example, I've hardcoded `TextLayer` as the only wrapped type, but we'd presumably want to wrap every instantiable type. When you call a type without `new`, you now get a protonope, as you might expect. (It always says line 5 though since that's where the `new Error()` call is in the wrap script -- not sure how to fix that easily.)

image


spicyj:
We could also prohibit `new` and do it implicitly if we wanted to make the code feel Swift-ier. Lots of JS libraries actually do the implicit instantiation thing.

saniul:
Looks great! Thanks, Ben!

I would vote for doing the implicit jnstantiation on user's behalf, since this is obviously what the user wanted to do anyway.

andymatuschak:
Ahaha way cool, @spicyj! Will dig into this in the next couple days (we're in NYC now for early math stuff hiiiiiiii)

andymatuschak:
My preference would be to throw an exception (as Ben's done) if the user leaves off the `new`. I don't think it's a good idea to create variant dialects of JS—people using Prototope-JS should be able to refer to ordinary (and widely-available) JS resources to understand their environment.

@nsfmc? @jbrennan? Have opinions on that?

I'll get to work on productionizing this. How does "commandeering" work on GH? Do I just open up another PR?

saniul:
Yup, another PR works

jbrennan:
> My preference would be to throw an exception (as Ben's done) if the user leaves off the `new`.

+1

spicyj:
As a JS person (but not a Prototope person), I would not find it out of place to have these constructors not use `new` at all. Lots of popular JS libraries have "constructors" that you call without new. Like jQuery has you write `var foo = $.Deferred();` to create a promise-like object. I think my personal preference is to forbid new.

spicyj:
(Yeah, just open a new PR.)

saniul:
Relevant: http://jfire.io/blog/2013/03/20/newless-javascript/

andymatuschak:
(actually, as awesome as it is that @spicyj has sent us a prototype, integrating it with DRY will require some work; deprioritizing)

andymatuschak:
I did a bunch of the work here: https://github.com/Khan/Prototope/tree/pr/47

PR #55, opened 2015-03-12
Upgrading Prototope to Xcode 8.3b3's Swift (breaks earlier builds)

andymatuschak:
The inscrutable `override` vs. `convenience` vs. whatever rules appear to have changed again.

### DIFF
--- a/Protoscope/ConsoleView.swift
+++ b/Protoscope/ConsoleView.swift
@@ -40,7 +40,7 @@ class ConsoleView: UIView {
 		textView.textStorage.deleteCharactersInRange(NSMakeRange(0, textView.textStorage.length))
 	}
 
-	override init() {
+	init() {
 		let blurEffect = UIBlurEffect(style: .Dark)
 		visualEffectView = UIVisualEffectView(effect: blurEffect)
 		visualEffectView.autoresizingMask = .FlexibleWidth | .FlexibleHeight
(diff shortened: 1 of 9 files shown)


Protoscope/ConsoleView.swift
```diff
@@ -40,7 +40,7 @@ class ConsoleView: UIView {
 		textView.textStorage.deleteCharactersInRange(NSMakeRange(0, textView.textStorage.length))
 	}
 
-	override init() {
```
NachoSoto: This is so weird. I had to fix it for [iOS too](https://github.com/Khan/iOS/commit/ee88c312d397bf995de108db1c5a744b932232f9). Why wouldn't this require `override` anymore?

Prototope/Animation.swift
```diff
@@ -161,7 +161,7 @@ public class Animator<Target: AnimatorValueConvertible> {
 		return (self.shouldAnimateLayer ? layer?.view.layer : layer?.view)
 	}
 
-	/** Immediately stops the animation. */
+	/** On a brighter note, with beta 3, our app now compiles -O for the first time ever! Hooray and congrats. */
```
NachoSoto: :heart:

andymatuschak: Ahahahaha wtf

Protoscope/ConsoleView.swift
andymatuschak: It is a mystery

jbrennan: My guess is because `init` is not a designated initializer of `UIView` but it's just a guess, and they didn't document it.

PR #56, opened 2015-03-14
Added a ScrollLayer.decelerationRetargetingHandler API

andymatuschak:
Because @may-li asked very nicely.

Bridging this without creating a retain cycle was… annoying.


saniul:
Looks good!

andymatuschak:
Thanks!

Issue #50, opened 2015-03-11
Insetting APIs for Rect


andymatuschak:
Would like to be able to do things like:

```
var someRect = new Rect({x: 5, y: 10, width: 20, height: 30})
var x = someRect.inset({top: 30, right: 15})
var y = someRect.inset({vertical: 30, horizontal: 15})
var z = someRect.inset(20)
```

jbrennan:
Wait, are the return values supposed to be rects or are they x/y/z coords?

jbrennan:
Wait, are the return values supposed to be rects or are they x/y/z coords?

jbrennan:
Wait, are the return values supposed to be rects or are they x/y/z coords?

NachoSoto:
I keep reading this as `Inserting APIs for React`.

spicyj:
Me too.

saniul:
> Wait, are the return values supposed to be rects or are they x/y/z coords?

I understood those to be rects

saniul:
Feedback wanted for https://github.com/Khan/Prototope/pull/58

jbrennan:
Fixed by #58

PR #58, opened 2015-03-16
[FIN] Insetting APIs for Rect

saniul:
Implements https://github.com/Khan/Prototope/issues/50

- [x] Swift API
- [x] JS Bridge
- [x] Documentation

### DIFF
--- a/Examples/TestInsets/main.js
+++ b/Examples/TestInsets/main.js
@@ -1,0 +1,23 @@
+/* 
+
+And example to demonstrate how to use insets on Rects
+
+*/
+
+
+var r1 = new Rect({x:100, y:100, width:250, height:250})
+
+console.log(r1.minX+","+r1.minY+":"+r1.maxX+","+r1.maxY);
+
+//r1 = r1.inset({value: 10})
+//r1 = r1.inset({horizontal: 20, vertical: 10})
+var r1 = r1.inset({top:10, right:20, bottom:30, left:40})
+
+// This will trigger a protonope
+//r1 = r1.inset({top:1000, right:20, bottom:30, left:40})
+
+
+console.log(r1.minX+","+r1.minY+":"+r1.maxX+","+r1.maxY);
+
+
+
--- a/PrototopeJSBridge/GeometryBridge.swift
+++ b/PrototopeJSBridge/GeometryBridge.swift
@@ -187,12 +187,14 @@ import Prototope
         let simplified = hasKey("vertical") || hasKey("horizontal")
         let core = hasKey("top") || hasKey("right") || hasKey("bottom") || hasKey("left")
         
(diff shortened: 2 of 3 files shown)


PrototopeJSBridge/GeometryBridge.swift
```diff
@@ -173,6 +174,31 @@ import Prototope
...
+        }
+        let getValue = { (key: String) -> Double in
+            return (args.objectForKey(key) as! Double?) ?? 0
+        }
+        
+        let justValue = hasKey("value")
+        let simplified = hasKey("vertical") || hasKey("horizontal")
+        let core = hasKey("top") || hasKey("right") || hasKey("bottom") || hasKey("left")
+        
+        switch (justValue, simplified, core) {
```
jbrennan: What do you think about using ifs here instead? This way, you can avoid the truth-table switch statement, which might make the intention a little clearer:

```swift
if let justValue = hasKey("value") {
    return RectBridge(rect.inset(value: getValue("value")))
}

if let simplified = hasKey("vertical") || hasKey("horizontal") {
    return ...
}
// ...

Environment.currentEnvironment?.exceptionHandler("Trying to call inset on with invalid parameters: \(args)")
return self
```

Examples/TestInsets/main.js
```diff
@@ -0,0 +1,23 @@
+/* 
+
+And example to demonstrate how to use insets on Rects
+
+*/
+
+
+var r1 = new Rect({x:100, y:100, width:250, height:250})
+
+console.log(r1.minX+","+r1.minY+":"+r1.maxX+","+r1.maxY);
```
andymatuschak: Ooh, that reminds me: we should override `toString()` on these things. Filed #59.

PrototopeJSBridge/GeometryBridge.swift
andymatuschak: Think this is a good suggestion. Reading the `case`s, I find myself having to look up at the switch expression to remind myself which is which.

andymatuschak:
Very sorry for the super long turnaround on this. API surface looks good!

saniul:
Don't apologise, I myself haven't really had time to work on any of the outstanding PRs this past week :(

PrototopeJSBridge/GeometryBridge.swift
saniul: Ah, the point here was to make sure that users of this API don’t try to mix and match the possible sets arguments.

andymatuschak: Ah, I see. I think that's worth doing: otherwise, they could end up with silent failures. :+1:

saniul:
I guess that’s it for this PR. gh-pages docs will come separately

andymatuschak:
woooo!

PR #60, opened 2015-03-24
Masking API

andymatuschak:
Will follow up with gh-pages docs pull request.

gif of feature

### DIFF
--- a/Prototope/Layer.swift
+++ b/Prototope/Layer.swift
@@ -374,6 +374,15 @@ public class Layer: Equatable, Hashable {
 		}
 	}
     
+    /** The mask layer is used to clip or filter the contents of a layer. Those contents will be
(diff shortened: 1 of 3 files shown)


Prototope/Layer.swift
```diff
@@ -373,6 +373,28 @@ public class Layer: Equatable, Hashable {
...
+
+        Be aware: mask layers do incur an additional performance cost. If the cost becomes too
+        onerous, consider making flattened images of the masked content instead. */
+    public var maskLayer: Layer? {
+        willSet {
+            newValue?.parent = nil
+            newValue?.maskedLayer?.maskLayer = nil
+        }
+        didSet {
+            view.maskView = maskLayer?.view
```
jbrennan: TIL

andymatuschak: I added this in iOS 8. :dancers:

PR #64, opened 2015-03-31
Shapes

jbrennan:
What it does
-----------------

- Adds a `ShapeLayer` type that supports the visual display of `Path` objects.
- Adds a `Path` type, a simplified wrapper for bezier paths for drawing lines and curves, either opened or closed
- More closely modelled on paper.js than on `UIBezierPath`.

To do
--------

- [ ] I'm still unclear what the relationship of the `frame` and the `bounds` should be. Specifically, if I create a path like `Path(rectangle: Rect(x: 50, y: 50, width: 10, height 10))` and then create a shape layer with that path, I kind of expect the layer to have an origin of `50, 50`, because that's the rectangle I gave when creating a path. But that kinda feels like a leaky abstraction too. What to do!
- [x] Path initializers aren't filled in yet. I could write these using the `Segment` type I made, but then I have to do some of the bezier math myself. I could instead just use `UIBezierPath` for those, but then I lose the ability to access segments / path points like if I'd just created the whole path manually. It's likely there's a simpler solution I'm just not seeing. Any ideas?
- [ ] JS Bridge


andymatuschak:
>  I'm still unclear what the relationship of the frame and the bounds should be. Specifically, if I create a path like Path(rectangle: Rect(x: 50, y: 50, width: 10, height 10)) and then create a shape layer with that path, I kind of expect the layer to have an origin of 50, 50, because that's the rectangle I gave when creating a path. But that kinda feels like a leaky abstraction too. What to do!

This is really unfortunate. Certainly the same dichotomy exists for `CGPath` / `CAShapeLayer`, but it's confusing and unpleasant there too. I think I didn't understand what you'd meant earlier about the frame/bounds issue when you'd mentioned it.

This seems like a significant vote for the idea that `ShapeLayer` and `Path` should be a single thing.

> Path initializers aren't filled in yet. I could write these using the Segment type I made, but then I have to do some of the bezier math myself. I could instead just use UIBezierPath for those, but then I lose the ability to access segments / path points like if I'd just created the whole path manually. It's likely there's a simpler solution I'm just not seeing. Any ideas?

You can access the points of a `UIBezierPath` by using `CGPathApply`.

jbrennan:
@andymatuschak I think this is good to go...Going to make separate issues for remaining tasks.

Prototope/Geometry.swift
```diff
@@ -80,6 +80,11 @@ public func *(scalar: Double, a: Point) -> Point {
     return a * scalar
 }
 
+
+public func*(a: Point, b: Point) -> Point {
```
andymatuschak: Space between `func` and `*` plz

andymatuschak: Probably worth documenting here that this is element-wise multiplication, not an outer product or something: there are a number of valid "multiplications" for points/vectors.

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+//  Prototope
+//
+//  Created by Jason Brennan on Mar-27-2015.
+//  Copyright (c) 2015 Khan Academy. All rights reserved.
+//
+
+import UIKit
+
+
+/** This layer represents a 2D shape, which is drawn from a supplied `Path` object. */
```
andymatuschak: There's no `Path` anymore!

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+		
+		self.shapeViewLayer.path = self.bezierPath.CGPath
+		self.shapeViewLayerStyleChanged()
+	}
+	
+	
+	// MARK: - Segments
+	
+	/** A list of all segments of this path. */
+	public var segments: [Segment]
```
andymatuschak: Don't you need `didSet { setNeedsDisplay() }` here?

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+		}
+		
+		self.segments[index] = segment
+		
+	}
+	
+	
+	// TODO(jb): How can this be triggered automatically when mutating the segments?
+	/** Redraws the path. You can call this after you change path segments. */
+	public func setNeedsDisplay() {
```
andymatuschak: Did you mean to make this public? Confused: are clients expected to always call this when they make a change?

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+	
+	
+	private class ShapeView: UIView {
+		override class func layerClass() -> AnyClass {
+			return CAShapeLayer.self
+		}
+	}
+	
+	
+	private func shapeViewLayerStyleChanged() {
```
andymatuschak: Nit: Cocoa style would be `shapeViewLayerStyleDidChange()`

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+		/** Lines will be joined with a miter style. */
+		case Miter
+		
+		/** Lines will be joined with a round style. */
+		case Round
+		
+		/** Line joins will have beveled edges. */
+		case Bevel
+		
+		internal func joinStyleString() -> String {
```
andymatuschak: `internal` is redundant here: it'll be `internal` by default.

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+		/** The line cap will have butts for ends. */
+		case Butt
+		
+		/** The line cap will have round ends. */
+		case Round
+		
+		/** The line cap will have square ends. */
+		case Square
+		
+		internal func capStyleString() -> String {
```
andymatuschak: `internal` is redundant here: it'll be `internal` by default.

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+			}
+			
+		}
+		for segment in self.segments {
+			drawSegment(segment)
+		}
+		
+		if self.closed && self.segments.count > 0 {
+			drawSegment(self.segments[0])
+			bezierPath.closePath()
```
andymatuschak: Isn't this unnecessary given that you've just drawn a segment back to the first segment?

Prototope/ShapeLayer.swift
jbrennan: This is what I'm not sure of. I made it public because if they mutate the list of segments on their own, then they need to inform we need a redraw (following paper's design here).

Could instead make this read only and only let you mutate through Approved Methods™ instead?

Prototope/ShapeLayer.swift
jbrennan: Good point!

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+				isFirstSegment = false
+			} else {
+				if let segmentHandleIn = segment.handleIn {
+					currentHandleIn = currentPoint + segmentHandleIn
+				} else {
+					currentHandleIn = currentPoint
+				}
+				
+				
+				if currentHandleIn == currentPoint && currentHandleOut == previousPoint {
```
andymatuschak: Gonna have to take your word for this logic: really can't understand what's going on here! Some comments about your strategy here would definitely be helpful.

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+			self.shapeViewLayer.strokeColor = strokeColor.CGColor
+		} else {
+			self.shapeViewLayer.strokeColor = nil
+		}
+		
+		self.shapeViewLayer.lineWidth = CGFloat(strokeWidth)
+	}
+	
+	
+	private var bezierPath: UIBezierPath {
```
andymatuschak: This is the kind of code that could really use some unit tests. Lots of edge cases here—what if there's only one segment on a closed path? What if there are no segments? What if we alternate straight and curved segments? Etc

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+/** Convenience functions for creating shapes. */
+extension Segment {
+	
+	// Magic number for approximating ellipse control points.
+	private static let kappa = 4.0 * (sqrt(2.0) - 1.0) / 3.0
+	
+	/** Creates a set of segments for drawing an oval in the given rect. Algorithm based on paper.js */
+	static func segmentsForOvalInRect(rect: Rect) -> [Segment] {
+		
+		let kappaSegments = [
```
andymatuschak: Whoa

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+			
+			segments.append(Segment(point: bottomRight - Point(x: 0.0, y: radius), handleIn: nil, handleOut: Point(x: 0.0, y: handle)))
+			segments.append(Segment(point: bottomRight - Point(x: radius, y: 0.0), handleIn: Point(x: handle, y: 0), handleOut: nil))
+			
+		}
+		return segments
+	}
+	
+	
+	/** Segments for a line. Algorithm based on something I just made up. */
```
andymatuschak: HOW DO YOU DO IT JASON

Prototope/ShapeLayer.swift
andymatuschak: Oh, yeah, I think it's definitely error-prone to rely on the client to tell you when something changed. It should be sufficient to add `didSet { setNeedsDisplay() }` to the `segments` definition above. That'll handle `addSegment`, etc. too.

Prototope/ShapeLayer.swift
```diff
@@ -0,0 +1,424 @@
...
+	
+	/** Initialize the ShapeLayer with a given path. */
+	public init(segments: [Segment], closed: Bool = false, parent: Layer? = nil, name: String? = nil) {
+		
+		self.segments = segments
+		self.closed = closed
+		
+		super.init(parent: parent, name: name, viewClass: ShapeView.self)
+		
+		self.shapeViewLayer.path = self.bezierPath.CGPath
```
andymatuschak: Call `setNeedsDisplay` instead of copying?

Prototope/ShapeLayer.swift
andymatuschak: Also this should probably actually defer the update: someone adding `n` segments in a single frame would have pathological performance if this method ran `n` times.

You can do that by using a view class like this:

```
class ShapeLayer: Layer {
  //...
  init(/*...*/) {
    view.displayHandler = { [weak self] in
      self?.shapeViewLayer.path = self.bezierPath.CGPath
    }
  }

  private func setNeedsDisplay() {
    view.setNeedsDisplay()
  }
  
  class CustomDisplayView: UIView {
    var displayHandler: (() -> Void)?

    override func setNeedsDisplay() {
      // The UIKit implementation (reasonably) won't call through to `CALayer` if you don't implement `drawRect:`, so we do it ourselves.
      layer.setNeedsDisplay()
    }
    
    override displayLayer(layer: CALayer!) {
      displayHandler?()
    }
  }
}
```

Prototope/ShapeLayer.swift
jbrennan: I'll NEVER TELL!

Prototope/ShapeLayer.swift
jbrennan: The truth is I don't really either, entirely. This is modelled on paper.js's implementation. The idea of having "in and out" control handles on a point is strange to me, because we effectively have to keep track of the previous control points when adding a curve. And if there are no control handles, then we have a line.

PR #68, opened 2015-04-06
Bridge the Point.slopeToPoint() function.

### LINKED ISSUE #66: Bridge Point.slope method

jbrennan:
Is it possible (or does it make sense) to bridge an optional type?

The slope function is optional for the case where the slope is infinite/undefined (i.e., there's a divide by zero). For the bridge I'm just returning 0 because it seems more sensible to do this than really worry about it too much. Thoughts?

[Fixes #66]

### DIFF
--- a/PrototopeJSBridge/GeometryBridge.swift
+++ b/PrototopeJSBridge/GeometryBridge.swift
@@ -19,6 +19,7 @@ import Prototope
 	static var zero: PointJSExport { get } // exported manually
 	init(args: NSDictionary) 
 	func distanceToPoint(point: PointJSExport) -> Double
+	func slopeToPoint(point: PointJSExport) -> Double
 	var length: Double { get }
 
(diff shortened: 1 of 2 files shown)


andymatuschak:
Nil is bridged to undefined, which seems appropriate in this context.

jbrennan:
Do you think that'd be practical for prototyping purposes? (I suppose this applies equally well to the Swift version as it does the bridge :))

andymatuschak:
Yeah, sounds very reasonable to me.

jbrennan:
> Yeah, sounds very reasonable to me.

Er, practical for it to be nil or practical for it to be 0?

andymatuschak:
Nil! 0 is out and out wrong.

jbrennan:
@andymatuschak Should be gtg

andymatuschak:
:+1:

PR #74, opened 2015-04-15
Protocaster can now remember a specified device and reconnect to it

jbrennan:
Select a device to reconnect to and Protocaster will do its best to reconnect to it when it re-appears.

It’ll even do so after a relaunch of Protocaster (but for now, we don’t remember the last prototype you sent it, so it’s not that useful.. who wants to add remembering that, too?)


andymatuschak:
Thanks, @jbrennan!

saniul:
I can add remembering the scene

andymatuschak:
<3

Issue #76, opened 2015-04-16
One protocoaster notification is probably enough


jbrennan:
![screen shot 2015-04-16 at 11 24 42 am](https://cloud.githubusercontent.com/assets/131923/7184447/46ce2dd6-e42b-11e4-9fab-6ee71b0dc5df.png)

PR #77, opened 2015-04-16
Don't show the same notification a billion times in a row

jbrennan:
c.f., #76


andymatuschak:
Great idea. This always stresses me out.

PR #78, opened 2015-04-16
Fix Layer.globalPosition's coordinate space conversion

andymatuschak:
`position` is expressed in the parent’s coordinate space, not local space, so we have to use the parent to convert.


jbrennan:
Good catch.

PR #79, opened 2015-04-17
[Hold] Transform JavaScript with Babel before running

spicyj:
This lets you use let/const, arrow functions, and more.

I updated all the examples that I could figure out how to run so they use let/const. Also threw in an arrow function for fun, though there weren't many short functions so using them elsewhere is more a matter of taste.


NachoSoto:
Nice :D

saniul:
@spicyj, awesome! :+1:

Examples/Color Puddles/main.js
```diff
@@ -1,45 +1,40 @@
 Layer.root.backgroundColor = new Color({hue: 0.5, saturation: 0.8, brightness: 0.3})
 
 // TODO: you currently can't attach touch handlers to the root layer. oops.
-var touchLayer = new Layer()
+const touchLayer = new Layer()
 touchLayer.frame = Layer.root.bounds
 
-var touchesToLayers = {}
-var touchLayers = []
+const touchesToLayers = new Map()
```
spicyj: Also threw in some ES6 goodness around this.

jbrennan:
Can you sell me on the benefits of using this? Debugging of prototypes is unfortunately pretty challenging and often the line number errors are a last bastion of helpfulness in those situations (I wouldn’t want to lose them).

andymatuschak:
:heart: :heart: @spicyj!!!

@jbrennan: It lets us use features from [JS ES6](https://babeljs.io/docs/learn-es6/) without waiting for JavaScriptCore to finish implementing them. Being able to use arrows would be freaking sweet! I believe @spicyj implemented this because I was complaining about JavaScript's *nutso* variable scope semantics (variables declared in statement blocks [like within a `for`] have function scope, not block scope!!), and `let` fixes that.

It looks like we could preserve line numbers by using https://github.com/mozilla/source-map/. Will try to take a look at that this weekend!

jbrennan:
I’m going to vote we hold on merging this until we can get line numbers working. I agree scope is whacko in JS but I think it’s more tolerable than having incomprehensibler errors. :neckbeard:

andymatuschak:
👍

spicyj:
^^ Now updated with source map support. :)

saniul:
👍❤️❤️❤️

andymatuschak:
dammit @spicyj :D

spicyj:
I figured that it would be easier for me to get the JS infra set up here… feel free to make changes from here if you like though. :)

andymatuschak:
Oh my gosh and you added stack traces to the protonopes too. This is so lovely! Arrow functions, better scope semantics, *and* better diagnostics! :heart: :ship:

andymatuschak:
how to github

PR #80, opened 2015-04-17
[FIN] Protocaster: Remember last selected path

### LINKED ISSUE #75: Protocaster should remember the last prototope you sent
Similar to #42 but for the projects. In this case, I don’t think we need any kind of additional settings UI, just remember the last one.

saniul:
Resolves #75


jbrennan:
Thanks a bundle!

PR #81, opened 2015-04-18
Sounds that are playing should keep playing even if you stop referencing them

andymatuschak:
I wanted to write JS code like:

```
new Sound({name: “foo”}).play()
```

But the garbage collector would occasionally collect the `Sound` and stop it from playing completely!

### DIFF
--- a/Prototope/Sound.swift
+++ b/Prototope/Sound.swift
@@ -75,5 +75,5 @@ public struct Sound: Printable {
 	}
 }
 
-private var playingAVAudioPlayers = Set<(AVAudioPlayer)>()
-private var playingAVAudioPlayerDelegates = Set<(Sound.AVAudioPlayerDelegate)>()
+private var playingAVAudioPlayers = Set<AVAudioPlayer>()
+private var playingAVAudioPlayerDelegates = Set<Sound.AVAudioPlayerDelegate>()
--- a/Prototope/Sound.swift
+++ b/Prototope/Sound.swift
@@ -39,12 +39,41 @@ public struct Sound: Printable {
 
 	public func play() {
 		player.currentTime = 0
+		if player.delegate == nil {
+			let delegate = AVAudioPlayerDelegate()
+			player.delegate = delegate
+			playingAVAudioPlayerDelegates.insert(delegate)
+		}
+		playingAVAudioPlayers.insert(player)
 		player.play()
 	}
 
 	public func stop() {
 		player.stop()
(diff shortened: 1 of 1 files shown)


Prototope/Sound.swift
```diff
@@ -39,12 +39,41 @@ public struct Sound: Printable {
...
+
+		@objc func audioPlayerDecodeErrorDidOccur(player: AVAudioPlayer!, error: NSError!) {
+			player.delegate = nil
+			playingAVAudioPlayers.remove(player)
+			playingAVAudioPlayerDelegates.remove(self)
+		}
+	}
 }
+
+private var playingAVAudioPlayers = Set<(AVAudioPlayer)>()
```
NachoSoto: Curious: why the parenthesis?

Prototope/Sound.swift
```diff
@@ -39,12 +39,41 @@ public struct Sound: Printable {
...
 	public func stop() {
 		player.stop()
+		if let delegate = (player.delegate as? Sound.AVAudioPlayerDelegate) {
+			playingAVAudioPlayerDelegates.remove(delegate)
+			player.delegate = nil
+		}
+		playingAVAudioPlayers.remove(player)
 	}
 
 	public static var supportedExtensions = ["caf", "aif", "aiff", "wav"]
```
NachoSoto: This can be `let` right? :)

Prototope/Sound.swift
andymatuschak: Oh, because I'm allocating an instance, not just declaring the type.

Prototope/Sound.swift
andymatuschak: Yep!

Prototope/Sound.swift
NachoSoto: Oh no, I meant around the type
```swift
Set<(AVAudioPlayer)>
```
instead of
```swift
Set<AVAudioPlayer>
```

andymatuschak: Oh, 'cause I'm a dummy!

andymatuschak: I think it's because I'd initially made it a single set of `(AVAudioPlayer, AVAudioPlayerDelegate)`, but that didn't compile because a tuple of `Hashable`s isn't necessarily `Hashable`. So I made two `Set`s, but left the parens around.

NachoSoto: Oh that makes sense :)

NachoSoto:
Looks good! :tada:

PR #86, opened 2015-05-18
[FIN] Bridged XXXAnimator.stop()

saniul:
We’d missed this one too.


saniul:
@jbrennan @andymatuschak

jbrennan:
Thanks @saniul !

PR #91, opened 2015-06-09
Reload protoscope when you press Cmd+R

jbrennan:
Previously, the only way to reload is by hitting save in your text editor (or changing another file). Now we can reload on demand. A little nicer than having to switch back between apps.


jbrennan:
(by the way, I do know how to merge branches myself, but I like doing Pull Requests because they let me give a little more context to *what* I’m doing and *why* I’m doing it)

Issue #94, opened 2015-06-15
I can haz custom fonts?


nefaurk:
It appears that bundling a font in your Prototope folder doesn't allow you to use that custom font in your prototype.

NachoSoto:
I think custom fonts need to be listed in the app's `plist`, so it might not be possible to load an arbitrary font at runtime.

saniul:
There's this: http://stackoverflow.com/questions/14048549/how-to-dynamically-load-a-font-under-ios-for-real

Seems like it's doable.

andymatuschak:
Weird! It won't let me assign you! But you have commit privileges here. Hmmmm

andymatuschak:
Ooooops, forgot to add the group I made to the collaborators list here. :D

PR #95, opened 2015-06-16
[FIN] Custom font support

saniul:
Resolves https://github.com/Khan/Prototope/issues/94

One quirk – the name of the font file should reflect the same name, that can be used by a `UIFont(name:size:)` call.

screen shot 2015-06-16 at 00 27 01


nefaurk:
Having trouble getting this working. I have a file named "Proxima Nova Regular.otf" in the same folder as my Prototope script, but it's still not picking it up when I set the fontName attribute to "Proxima Nova Regular". Thoughts on what I'm doing incorrectly?

jbrennan:
@nefaurk Have you added it to the info plist? I think that’s necessary?

saniul:
> Have you added it to the info plist? I think that’s necessary?

@jbrennan not necessary, this approach allows us to load the font files dynamically.

Prototope/FontProvider.swift
```diff
@@ -0,0 +1,76 @@
...
+		if let font = UIFont(name: name, size: CGFloat(size)) {
+			return font
+		}
+		
+		if let customFontData = resourceForFontWithName(name) {
+			let URL = NSFileManager.defaultManager().URLsForDirectory(NSSearchPathDirectory.CachesDirectory, inDomains: NSSearchPathDomainMask.UserDomainMask).first as! NSURL
+			
+			let fontFileURL = URL.URLByAppendingPathComponent(name)
+			
+			customFontData.writeToURL(fontFileURL, atomically: true)
```
saniul: Possible improvement – use `CTFontManagerRegisterGraphicsFont` [like Marco does](http://www.marco.org/2012/12/21/ios-dynamic-font-loading) instead of writing the data to a file and calling `CTFontManagerRegisterFontsForURL`.

PR #98, opened 2015-06-17
Adds touch dots to the app for recording where touches are.

jbrennan:
A Swift re-write of https://github.com/adamwulf/ios-uitouch-bluedots.

This adds phoney-finger dots for all touches when enabled. It shouldn’t interfere with any other touches or gesture recognizers in the app, so you can turn it on at will. This lets you record where your touches are.


Prototope/Screen.swift
```diff
@@ -0,0 +1,224 @@
...
+				removeTouchDots()
+			}
+		}
+	}
+	
+	private static var overlayView: TouchDotOverlayView? = nil
+	private static func enableTouchDots() {
+		overlayView = TouchDotOverlayView(frame: CGRect(Layer.root.frame))
+		let window = UIApplication.sharedApplication().delegate?.window
+		window!!.addSubview(overlayView!)
```
andymatuschak: The double-force is a little weird! Maybe it'd be better to use `delegate!` on the previous line—I think that makes your intent clearer (i.e. "this code assumes that the delegate is non-nil and also that the delegate's window is non-nil")

Prototope/Screen.swift
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+	
+	
+	func touchesEnded(touches: Set<UITouch>) {
+		for touch in touches {
+			self.removeViewForTouch(touch)
+		}
+	}
+
```
andymatuschak: Gotta implement `Cancelled` too!

andymatuschak: Oh, oops, I see how you route `cancelled` to `ended` below. Never mind!

Prototope/Screen.swift
```diff
@@ -0,0 +1,224 @@
...
+/** Internal gesture recognizer class to handle where touches are and how they interact with other GR in the system. */
+class TouchDotGestureRecognizer: UIGestureRecognizer, UIGestureRecognizerDelegate {
+	
+	var activeTouches = Set<UITouch>()
+	weak var touchDelegate: TouchDotGestureRecognizerDelegate?
+
+	init() {
+		
+		// TODO(jb): I'm not sure how to hack around this, I don't intend on messaging the target, but I've got to initialize super...
+		super.init(target: NSObject(), action: "description")
```
andymatuschak: I guess I don't quite understand why you use a custom delegate protocol instead of `UIGestureRecognizer`'s built-in target/action machinery!

andymatuschak:
Until we bother writing some UI to activate/deactivate this, maybe put a conspicuous bool at the top of Protoscope's app delegate to turn this on/off?

Prototope/Screen.swift
jbrennan: Only because I was basically re-writing the ObjC source I linked to and that’s what it did (and didn’t feel like spending too much time futzing with it). But I can go ahead and write it the better way you suggest if you’d like.

andymatuschak: Nah, s'cool

PR #99, opened 2015-06-22
Integrate Early Math Prototypes into a player app

jbrennan:
What it does
-----------------

This pull request integrates prototypes from the Early Math Prototypes repo (as a submodule) into a local player app so they can be played on the iPad without needing to be tethered / served over wifi.

- Handles loading JS prototypes
- Lists all available prototypes
- handles images and sounds

Notes
------

Everything should be functional but I want to clean up the code style a little and add a few docs here or there. I’d love to have feedback on this (and spotting for any showstoppers) but it should be more or less ready to go.


Protomath/PrototypeProvider.swift
```diff
@@ -0,0 +1,107 @@
...
+				Prototype(directoryURL: NSURL(fileURLWithPath: earlyMathPath + "/" + $0, isDirectory: true)!, name: $0)!
+			}
+		} else {
+			self.prototypes = []
+		}
+		
+	}
+	
+	
+	func numberOfPrototypes() -> Int {
```
andymatuschak: Why these three methods instead of having clients use `prototypes`? i.e.

```
provider.prototypes.count // numberOfPrototypes()
provider.prototypes[index].name // prototypeNameAtIndex
provider.prototypes[index].mainFileURL // pathForPrototypeAtIndex

File: Protomath/PrototypeProvider.swift [hunk 2]
```diff
@@ -0,0 +1,107 @@
...
+}
+
+
+struct Prototype {
+	let mainFileURL: NSURL
+	let name: String
+}
+
+
+extension Prototype {
```
andymatuschak: Mind adding a copy pasta notice here, since this is copied from `Prototype.swift`? Or can you share that code?

andymatuschak:
This is very exciting! One broad thought: since Prototope is more broadly-useful OSS, it feels odd to add a target for Early Math to its Xcode project.

Would you be open to making a separate repository which references this one as a submodule and builds the Prototope framework as a subproject?

jbrennan:
> Would you be open to making a separate repository which references this one as a submodule and builds the Prototope framework as a subproject?

Yeah absolutely. I thought of this but decided to just go with the existing project only because I don’t have the rights to make repos for KA. But I’ll whip one up today and then worry about that later.

Protomath/PrototypeProvider.swift
jbrennan: Originally it was for info hiding / separation of concerns / wasn’t sure what `Prototype` looked like but yeah, probably makes more sense to just punt down the object.

jbrennan:
> references this one as a submodule and builds the Prototope framework as a subproject?

Seems like it might be cleaner just to have the Early Math Prototype Player project have Prototope has a submodule/project then for it to have OHai.. ?

andymatuschak:
Yep

jbrennan:
https://github.com/jbrennan/Early-Math-Prototype-Player For now.

PR #105, opened 2015-09-16
Experimental Mac support

jbrennan:
Hullo hullo! Over the past little while I’ve been working away at experimental support for prototope on the Mac. It works quite nicely, but there are lots of little issues.

My main goal isn’t for the Mac port to be perfect, but for it to:

1. Work well enough to do basic prototyping in.
2. Not make the iOS code significantly crappier.

I think I’ve achieved both goals. Most of the work was done in existing iOS-based files, but #def’d out for specific OS X stuff.

Done
-------

- Basic support for layers and the layer hierarchy
- Dealing with sounds, speech, and images
- video support (not super tested)
- Camera support (not really tested)
- Mouse events

Not really done yet, but maybe some day
------

- JS bridge
- Text layers
- Scroll layers
- Keyboard events
- Multiple windows
- lots of todos

Also of note
-------

I had significant trouble getting this into a reusable OS X framework. Frameworks don’t seem to like including other frameworks or using bridging headers (although, much chagrin, the iOS framework *somehow* does this). I wasted about three working days on it over the course of the project.

The “fix” was to make my own fork of `pop`, create a static library target for OS X for it, and then point the Prototope OS X framework at that. I should probably try to get my changes merged upstream with Facebook, or figure out what the hell is the problem with it.


Prototope/Animation.swift
```diff
@@ -70,22 +73,41 @@ public class LayerAnimatorStore {
...
+                (obj as! SystemView).frame.origin.x = values[0]
+                (obj as! SystemView).frame.origin.y = values[1]
             }
         } as! POPAnimatableProperty
         origin = Animator(layer: layer, property: animatableOrigin)
         
+		#if os(iOS)
 		position = Animator(layer: layer, propertyName: kPOPLayerPosition)
+			#else
+			// I suspect all the kPOPLayer* properties should *not* be used on our Layer's UIView, but instead its CALayer.
```
andymatuschak: Yeah, I think that's true. Might as well do it on both platforms?

Prototope/Color.swift
```diff
@@ -6,11 +6,18 @@
...
 
-import UIKit
+#if os(iOS)
+	import UIKit
+	typealias SystemColor = UIColor
+#else
+	import AppKit
+	typealias SystemColor = NSColor
+#endif
+
```
andymatuschak: Neat!

Prototope/Color.swift
```diff
@@ -50,23 +57,23 @@ public struct Color {
 	}
 
 	/** Constructs a Color from a UIColor. */
-	init(_ uiColor: UIColor) {
+	init(_ uiColor: SystemColor) {
```
andymatuschak: Probably worth changing this argument and property name to, like, `systemColor`.

Prototope/DisplayLink.swift
```diff
@@ -0,0 +1,88 @@
+//
+//  DisplayLink.swift
+//  Prototope
+//
+//  Created by Jason Brennan on 2015-08-10.
+//  Copyright (c) 2015 Khan Academy. All rights reserved.
+//
+
+import AppKit
```
andymatuschak: Probably worth a big comment at the top that's like, "This file is OS X only!!!111"

Prototope/DisplayLink.swift
```diff
@@ -0,0 +1,88 @@
...
+
+
+typealias HeartbeatDisplayLinkCallback = (sender: SystemDisplayLink) -> Void
+
+/** Crappy wrapper around CVDisplayLink to act pretty close to a CADisplayLink. Only OS X kids get this. */
+class DisplayLink: NSObject {
+
+	private let displayLink:CVDisplayLink? = {
+		var linkRef:Unmanaged<CVDisplayLink>?
```
andymatuschak: Spaces between identifier and type here and on the previous line please~~~

Prototope/DisplayLink.swift
```diff
@@ -0,0 +1,88 @@
...
+			_:CVDisplayLink!,
+			_:UnsafePointer<CVTimeStamp>,
+			_:UnsafePointer<CVTimeStamp>,
+			_:CVOptionFlags,
+			_:UnsafeMutablePointer<CVOptionFlags>,
+			_:UnsafeMutablePointer<Void>)->Void in
+			
+			heartbeatCallback(sender: self)
+		}
+		self.dynamicType.DisplayLinkSetOutputCallback(self.displayLink!, callback: callback)
```
andymatuschak: Whoa interesting! I haven't actually seen anyone use `dynamicType` before. Why did you decide to use that here instead of saying `DisplayLink.DisplayLinkSetOutputCallback…`?

Prototope/DisplayLink.swift
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+}
+
+
+// Junk related to wrapping the CVDisplayLink callback function.
+extension DisplayLink {
+	private typealias DisplayLinkCallback = @objc_block ( CVDisplayLink!, UnsafePointer<CVTimeStamp>, UnsafePointer<CVTimeStamp>, CVOptionFlags, UnsafeMutablePointer<CVOptionFlags>, UnsafeMutablePointer<Void>)->Void
+	
+	private class func DisplayLinkSetOutputCallback(displayLink:CVDisplayLink, callback:DisplayLinkCallback) {
```
andymatuschak: Idiomatic naming for class/static methods in Swift would still be something like `setOutputCallback(displayLink: CVDisplayLink, ...)`

Prototope/DisplayLink.swift
```diff
@@ -0,0 +1,88 @@
...
+		let block:DisplayLinkCallback = callback
+		let myImp = imp_implementationWithBlock(unsafeBitCast(block, AnyObject.self))
+		let callback = unsafeBitCast(myImp, CVDisplayLinkOutputCallback.self)
```
andymatuschak: Hahahaha oh maaaaaaan so sweet

Prototope/InputEvent.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  InputEvent.swift
+//  Prototope
+//
+//  Created by Jason Brennan on 2015-08-17.
+//  Copyright (c) 2015 Khan Academy. All rights reserved.
+//
+
+import AppKit
```
andymatuschak: Ditto comment about top-level "OS X only!" docstring

Prototope/Layer.swift
```diff
@@ -219,8 +255,22 @@ public class Layer: Equatable, Hashable {
 	/** The origin and extent of the layer expressed in its parent layer's coordinate space.
 		Animatable. */
 	public var frame: Rect {
-		get { return Rect(layer.frame) }
-		set { layer.frame = CGRect(newValue) }
+		get {
+			// TODO(jb): Do we need to make this distinction? Can't UIKit's version just use view.frame instead of layer.frame?
```
andymatuschak: Oh. Yeah, that seems fine.

Prototope/Layer.swift
```diff
@@ -459,7 +516,20 @@ public class Layer: Equatable, Hashable {
 			return emitter !== particleEmitter
 		}
 	}
+	
+	
+	// TODO(jb): Port touches / gestures to OS X? What makes sense here?
```
andymatuschak: There's `NSGestureRecognizer` now!

Prototope/Layer.swift
```diff
@@ -637,52 +760,72 @@ public class Layer: Equatable, Hashable {
...
 		required init(coder aDecoder: NSCoder) {
 			fatalError("This method intentionally not implemented.")
 		}
 
 		override init(frame: CGRect) {
 			super.init(frame: frame)
+			#if os(OSX)
+				self.wantsLayer = true
+				
+				// TODO(jb): Can probably lazily add this when a mouse Entered/moved/exited event happens, so not all layers need to have tracking areas by default.
```
andymatuschak: What about adding this when a `mouseMovedHandler` is set? My understanding is that these are quite expensive.

Prototope/Layer.swift
```diff
@@ -820,3 +1007,47 @@ private func incorporateTouches(touches: NSSet, intoTouchSequenceMappings mappin
...
+#if os(iOS)
+	import UIKit
+	public typealias SystemImageView = UIImageView
+	#else
+	import AppKit
+	public typealias SystemImageView = NSImageView
+	
+	extension SystemView {
+		func setNeedsLayout() {
+			// TODO(jb): What's the OS X equiv of this again?
```
andymatuschak: It's `setNeedsLayout(true)`.

Prototope/Layer.swift
```diff
@@ -820,3 +1007,47 @@ private func incorporateTouches(touches: NSSet, intoTouchSequenceMappings mappin
...
+			// no-op?
+		}
+		
+		func layoutSubviews() {
+			self.resizeSubviewsWithOldSize(self.frame.size)
```
andymatuschak: Why is this necessary? Isn't this what would normally happen?

Prototope/Layer.swift
```diff
@@ -820,3 +1007,47 @@ private func incorporateTouches(touches: NSSet, intoTouchSequenceMappings mappin
...
+				if let color = self.layer?.backgroundColor {
+					return SystemColor(CGColor: color)
+				}
+				return nil
+			}
+			set {
+				if let systemColor = newValue {
+					self.layer?.backgroundColor = systemColor.CGColor
+				} else {
+					self.layer?.backgroundColor = SystemColor.clearColor().CGColor
```
andymatuschak: Should this be set to `nil` instead?

andymatuschak:
This is super ultra cool!!!~~~

Prototope/DisplayLink.swift
jbrennan: Yeah, this is a mess. For now I’d like to leave it as is because I’m going to need to re-write it for Swift 2. I’ve left a note that the class is currently kind of busted.

Prototope/DisplayLink.swift
jbrennan: Definitely. Will rename (most of this class is derived from a Stack Overflow answer..)

This class can be made way nicer with Swift 2’s support for function pointers.

Prototope/Layer.swift
jbrennan: Yeah, I’ve played around with those a bit before, but wasn’t sure if that’d really make sense for prototyping OS X apps. I’ll probably give it a shot eventually though!

Prototope/Layer.swift
jbrennan: I think that was because `NSView` doesn’t have `layoutSubviews`?

Prototope/Layer.swift
jbrennan: Hmm, it appears `NSView` doesn’t have a method called `setNeedsLayout(_:Bool)`

jbrennan:
Wow, so I left this for a looooooong time, sorry about that.

It’s a little hard to test right now because this branch is Swift 1.2 and I don’t have any old Xcodes kicking around. But I think it’s solid enough (the OS X prototope is experimental anyway). I plan on fixing up the Swift 2 branch as soon as this gets merged in.

Prototope/DisplayLink.swift
andymatuschak: Okeedoke!

Prototope/Layer.swift
andymatuschak: There's no argument! It's just `NSView.setNeedsLayout()`

andymatuschak: Oh, wait, no, that's `CALayer` that has no argument. I'm confused now because Dash does claim that `NSView` has such a method (introduced in 10.7) but developer.apple.com doesn't??

Prototope/Layer.swift
andymatuschak: Right!

andymatuschak:
Alright, whatevs, let's get this going! :D

PR #108, opened 2015-10-01
Swift 2

jbrennan:
Upgrades Prototope to Swift 2.

(This should be looked at after #105 is merged in)

Issue #41, opened 2015-03-05
Our Layer subclasses are missing touch handling


andymatuschak:
`Layer`'s view is a special `UIView` subclass which has some additional features—most notably, key parts of the `touchBeganHandler` implementation. As a consequence, touch handling doesn't work on e.g. text layers, video layers, etc.

The `Layer` subclasses all use their own view subclasses. There's no multiple inheritance, so we'll need to use composition instead (i.e. make those custom views be a subview of `Layer`'s special view), not *be* the layer's view.

/cc @jbrennan

jbrennan:
Deuce.

PR #109, opened 2015-12-20
Image from text

jbrennan:
This patch lets you create an `Image` by rendering a given string (with optional `font` and `textColor` parameters) into a buffer.

This is useful if you need to make an image from text (in particular, emoji) for things like particles, or if you need a text layer and you want to add touch handling to it, but you can’t because of #41 

This should be looked at after Swift 2 changes are merged in (#108)

### DIFF
--- a/Prototope.xcodeproj/project.pbxproj
+++ b/Prototope.xcodeproj/project.pbxproj
@@ -1440,7 +1440,7 @@
 		EB306C0119DF77CC00113FA3 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftMigration = 0700;
+				LastSwiftMigration = 0720;
 				LastSwiftUpdateCheck = 0700;
 				LastUpgradeCheck = 0600;
 				ORGANIZATIONNAME = "Khan Academy";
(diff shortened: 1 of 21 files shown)


andymatuschak:
Whoa, some rough Git stuff happened here. I think 68a4206 is the only relevant revision?

jbrennan:
> Whoa, some rough Git stuff happened here. I think 68a4206 is the only relevant revision?

Yeah, I think it’s because I was branching off of a branch off a branch, etc. As far as git is concerned, none of these commits were on master. I’ll try fixing this up (I imagine the Swift 2 PR suffers the same problems).

jbrennan:
> Thought: for the same reason it's useful to be able to get an image of text, it's useful to be able to get an image of anything a layer can draw (e.g. a bezier path, etc). What if you could get an image (i.e. a snapshot) from any layer?

Yeah, that’d be lovely. I’ll make an issue for that.

> Even if you could do that, you'd probably still want a convenience API that looks like this, so I'm cool with rolling with this too.

Sweet!

> Mind bridging this to JS? I don't want the API surface to get skewed.

Ah yes, I will do that.

jbrennan:
Ugh, more git messiness (I need a better strategy for keeping my fork in sync with Khan/Prototope). The relevant commit for the JS bridge is 1a080394 @andymatuschak

andymatuschak:
Sweet, thanks. :)

Issue #103, opened 2015-09-15
3D Touch support


jbrennan:
I don’t know what this would look like, but it’d probably be nice to have. I’m opening this to start a discussion on it!

wave2588:
I want to learn  3D Touch

nsfmc:
were you thinking of exposing the `force` property in `TouchSample` or a simplified way of throwing peek onto a layer?

jbrennan:
@nsfmc Oh good point! I suppose probably both, if that makes sense.

saniul:
I’d start by just adding the `force` property.

nsfmc:
:dancers:

PR #112, opened 2016-03-15
Apple Pencil + 3D touch support

### LINKED ISSUE #104: Apple Pencil support
I have no idea how it works, so maybe we’ll get a lot of this for free. But I want to play with it, so I want to make note of it :pencil2:

jbrennan:
This adds some basic Apple Pencil + 3D touch support to `TouchSample`, namely:

- `preciseGlobalLocation`, which gives very accurate touch locations for stylus touches
- `force`, which gives the pressure of the Apple Pencil touches. I think this also works for 3D touch but I don’t have a device to test on

The properties are also bridged to JS.

This PR also bumps the deployment targets for both Prototope and the JS bridge to `9.1` to take advantage of the new API.

In the future, it might make sense to give stylus input its own Sample type (for things like angles, estimated precise locations, etc).

fixes #104 and #103

### DIFF
--- a/Prototope/Gesture.swift
+++ b/Prototope/Gesture.swift
@@ -23,14 +23,14 @@ public struct TouchSample: SampleType {
 	
 	
 	/** The force of the touch. Available only for certain devices (3D touch devices like iPhone 6S or Apple Pencil). */
-	public let force: Double
+	public let force: Double?
 
 	/** The location of the touch sample, converted into a target layer's coordinate system. */
 	public func locationInLayer(layer: Layer) -> Point {
 		return layer.convertGlobalPointToLocalPoint(globalLocation)
 	}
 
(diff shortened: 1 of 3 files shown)


jbrennan:
screen shot 2016-03-14 at 11 08 08 pm

Prototope/Gesture.swift
```diff
@@ -14,18 +14,27 @@ import UIKit
...
+	
+	/** The precise location of the touch sample in the root layer's coordinate system. This should only be used for Stylus, not finger, input. */
+	public let preciseGlobalLocation: Point
 
 	/** The time at which the touch arrived. */
 	public let timestamp: Timestamp
+	
+	
+	/** The force of the touch. Available only for certain devices (3D touch devices like iPhone 6S or Apple Pencil). */
+	public let force: Double
```
andymatuschak: Should probably be optional, right?

PR #113, opened 2016-03-31
Fixed Rect.center setter

NachoSoto:
Notice while inspecting the code.


spicyj:
I think this is correct as-is? If you do `rect.center = rect.center`, it should add `0`, not set to `0`.

NachoSoto:
Hmm I'm not sure I understand. If you're setting the `center` of a `Rect` to a **new** value, the existing `origin` shouldn't matter?

spicyj:
If you have a 2x2 box with origin at (3, 4) then its center is at (4, 5). If I do `.center = (4, 6)` it should move the by over 1 unit. One way to do that is to adding (0, 1) to the origin, which you can calculate by subtracting (4, 6) – (4, 5).

spicyj:
```
origin = oldOrigin + (newCenter - oldCenter)
```

is equivalent to

```
origin = newCenter + (oldCenter - oldOrigin)
```

if that helps you think about it, where `oldCenter - oldOrigin` is a vector pointing from the origin to the center (half the box's size in each dimension).

andymatuschak:
(thanks, @spicyj—that is indeed what I intended!)

jbrennan:
(just a meta note: boy is this conversation hard to follow with only symbolic math; I wish there was an easy way for us to discuss (or even program!) these things geometrically)

NachoSoto:
Ooooh that makes a lot of sense, thanks for the explanation @spicyj. I don't know why I found it so unintuitive at first!

